PR #22, opened 2013-05-31
Re-enable metadata

david-macmahon:
This is a simplistic Hash parser, but it works for basic metadata
purposes which seems better than having no metadata support at all.  The
following examples show how to set a page title using metadata.

All on one line:

``` ruby
<!-- --- title: My Fantastic Page -->
```

Or split across multiple lines:

``` ruby
<!-- ---
title: My Fantastic Page
-->
```

Notes:
1. Only one key/value pair per line. Multiple lines must be used for
   multiple key/value pairs (though I'm not aware of support for any keys
   other than "title").
2. Keys and values cannot contain embedded newlines.
3. The patch does _not_ HTML-sanitize the key or value portion of the
   metadata.  Presumably metadata is not solely for inclusion in HTML, so
   only the user of the metadata will know whether it needs to be sanitized
   in any way.

### DIFF
--- a/lib/gollum-lib/markup.rb
+++ b/lib/gollum-lib/markup.rb
@@ -650,9 +650,10 @@ module Gollum
     #
     #########################################################################
 
-    # Extract metadata for data and build metadata table. Metadata
-    # is content found between markers, and must
-    # be a valid YAML mapping.
+    # Extract metadata for data and build metadata table.  Metadata consists of
+    # key/value pairs in "key:value" format found between markers.  Each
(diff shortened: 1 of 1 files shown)


gerwitz:
For the reference of anyone else who ends up on this issue looking for metadata handling, we have been happy using a simply YAML filter: https://github.com/TheArtificial/rails-wiki/blob/b640b797805bea93a643c7d2201f282cb3294a65/lib/frontmatter.rb

PR #77, opened 2014-04-06
Add caching adapter

sirupsen:
In a bigger deployment of Gollum, we've seen very slow performance from parts of Gollum. This adds the ability to pass a cacher that handles this. In our deployment, we use Memcached to share this state across workers.

@fw42 @camilo can you guys take a look at this too?

### DIFF
--- a/lib/gollum-lib/page.rb
+++ b/lib/gollum-lib/page.rb
@@ -217,7 +217,7 @@ module Gollum
     # Returns the String data.
     def formatted_data(encoding = nil, include_levels = 10, &block)
       if @blob
-        @doc ||= Gollum.cache.fetch("gollum:page:doc:#{Page.cname(self.path).downcase}") {
+        @doc ||= Gollum.cache.fetch("gollum:page:doc:#{page_cache_suffix}") {
           markup_class.render(historical?, encoding, include_levels)
         }
         return yield(@doc) if block_given?
@@ -473,8 +473,6 @@ module Gollum
       key = "gollum:page:sub:#{page_cache_suffix}"
 
       Gollum.cache.fetch(key) do
-        puts "find_sub_page: cache miss for #{key}"
-
         dirs = self.path.split('/')
         dirs.pop
         map = @wiki.tree_map_for(@wiki.ref, true)
@@ -499,6 +497,8 @@ module Gollum
     def inspect
       %(#<#{self.class.name}:#{object_id} #{name} (#{format}) @wiki=#{@wiki.repo.path.inspect}>)
     end
+
+    private
 
     def page_cache_suffix
       Page.cname(self.path).downcase


Naatan:
It's really a shame that this got closed and the devs just let it, this is a significant contribution that got hung up on some technicalities from my impression. Gollum-lib is awesome, but it's -slow-. It needs caching to be used in any production environment.

dometto:
@Naatan, thanks for pointing out this PR. Although I think the Cache adapter implementation is nice and clean, and could potentially be used to cache more stuff, note that the most important potential performance increase has already been implemented in [this commit](https://github.com/gollum/gollum-lib/commit/8167050de7fd65d635f129ae5a058684d6d7753a), which implements caching for `Page#formatted_data`.

I think at the moment the highest performance increase for gollum-lib can be gained from rewriting the tag filter. See discussion [here](https://github.com/gollum/gollum/issues/839).

Naatan:
Good to hear the devs are giving performance the needed attention. Note that I tried the caching code in this PR and noticed a significant performance improvement on restructured text pages in particular. So it seems the commit you referenced does not manage to cache certain information properly. I'm still having a bit of trouble reading Ruby (so much magic going on) but shouldn't it be yielding `@formatted_data`?

dometto:
@Naatan, that's interesting. I think @etienneCharignon  and @jonathanperret's caching of `#formatted_data` is working correct -- as the tests show, it should yield `doc` and return `@formatted_data`. If the cache implementation in this PR still improves performance, there must be some other part that's doing it, possibly the improvements to `Page#find_sub_page`?

Naatan:
Actually I ended up dumping the find_sub_page code as it was being troublesome in merging it into the latest version of gollum-lib and it seemed insignificant. I'll try and run some more tests once I have this code stabilized.

Naatan:
I think the issue is that gollum-lib stores the information on the class instance, and gollum does not preserve this instance for followup requests to the same page, it creates a new instance (assumption). So @formatted_data would be unset for each followup request.

Want me to create a new bug for this? We're getting fairly off-topic.

dometto:
Ah, that seems right. You could open a separate issue yes. Perhaps the best thing would be to see if we can use some kind of pre-existing cache gem, so we don't have to handle things like releasing caches ourselves.

Sirupsen:
If anyone is interested in picking this up, that's cool with me, but I've since lost complete context. We just ended up putting the thing on a RAM-based filesystem. The problem is that there are tons of `O(N)` IO calls all over IIRC.

Naatan:
That's an interesting approach, how big were the performance gains from that?

Sirupsen:
Massive, I think we went from 2-3s, page loads with probably a few thousand pages, to ~100ms. It's a super dirty ops approach though.

PR #128, opened 2014-12-30
Extract Org-specific code blocks. Resolves gollum#256.

dometto:
`gsub` org-specific code blocks in the code filter so that a block's contents won't be subjected to tag filtering. See https://github.com/gollum/gollum/issues/256. It's late, so I'll leave this open and would appreciate a quick glance. :)

### DIFF
--- a/lib/gollum-lib/filter/code.rb
+++ b/lib/gollum-lib/filter/code.rb
@@ -5,46 +5,42 @@
 # Render a block of code using the Rouge syntax-highlighter.
 class Gollum::Filter::Code < Gollum::Filter
   def extract(data)
-    return data if @markup.format == :txt
+    case @markup.format
+      when :txt
+        return data
+      when :asciidoc
+        data.gsub!(/^(\[source,([^\r\n]*)\]\n)?----\n(.+?)\n----$/m) do
+          cache_codeblock($2, $3)
+        end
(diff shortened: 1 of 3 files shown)


dometto:
According to [the syntax highlighting section in the wiki](https://github.com/gollum/gollum/wiki#syntax-highlighting), three-tilde style code blocks are a special gollum-wide (markup-independent) feature. So I propose that we extract those code blocks no matter what markup is being used, while we extract tilde-style code blocks only if `@markup.format == :markdown`. I think that could properly be treated as a bugfix rather than as a breaking change, even though people *could* currently be using the tilde-style code blocks outside of markdown. Thoughts?

bartkamphorst:
I thought vanilla markdown uses indentation for code blocks, and GFM uses the three-backticks thing. So I guess the tildes are gollum-specific?

dometto:
Ah, yes. I was confused because [kramdown](http://kramdown.gettalong.org/syntax.html#fenced-code-blocks) does support the tilde-syntax, but as they note, that's an addition to vanilla markdown. What do you think then? Perform the tilde and backtick substitutions for every markup? In that case we should add the tilde-syntax to the wiki.

dometto:
Pushed asciidoc tests.

PR #130, opened 2015-01-03
Reverse Creole link parts and refactor reversals. Resolves gollum#78.

dometto:
1. Refactor logic for reversing link parts
2. Also reverse Creole links (see https://github.com/gollum/gollum/issues/78)

(Reopened this against `rc`, we should do a minor-level release in the near future anyway.)


bartkamphorst:
Good improvement. :+1: And I agree with you about merging with `rc` instead of `master`.

PR #134, opened 2015-01-13
Experimental support for PlantUML (Do not merge).

hsanson:
Incomplete and experimental support PlantUML diagrams:  http://plantuml.sourceforge.net/preprocessing.html. 

Is in "works-for-me" state and this pull request is only to get feedback and request help on some  implementation questions:

  - How to provide gollum with details on where to find the plantuml.jar file?
  - How to provide gollum with the path for the java binary?
  - The plantuml.jar file generates image files (.png). Where should these images be placed?

My ultimate goal is to have gollum (as used in gitlab) accepting PlantUML blocks and replacing them with the generated images locally on the same server without relying on third party services like websequencediagrams. 

As additional benefit PlantUML supports more diagram types

### DIFF
--- a/lib/gollum-lib/filter/plantuml.rb
+++ b/lib/gollum-lib/filter/plantuml.rb
@@ -1,0 +1,52 @@
+# ~*~ encoding: utf-8 ~*~
+require 'net/http'
+require 'uri'
+require 'open-uri'
+
+# PlantUML Diagrams
+#
+# Render an inline plantuml diagram by generating a PNG image using the
+# plantuml.jar tool.
+#
+class Gollum::Filter::PlantUML < Gollum::Filter
+
+  JAR = "/home/ryujin/plantuml.jar"
+  JAVA = "/home/ryujin/Apps/jdk/bin/java"
+
+  # Extract all sequence diagram blocks into the map and replace with
+  # placeholders.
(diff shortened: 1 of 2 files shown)


dometto:
As I said in https://github.com/gollum/gollum/issues/931, we would prefer not to include (and support) something that depends on an external binary. So I think this particular feature will not be merged into gollum-lib (even in the future). If you want to help incorporate an alternative to WSD, that would be really great, but we'd rather look in another direction. Like I said, we would prefer a UML solution that is pure Ruby or JS based. I don't personally use anything like WSD/UML, but a quick search reveals that there is e.g. http://jumly.tmtk.net (and many others). 

That being said, if you want PlantUML support for personal use, you could of course provide a fork/patch of the project.

> How to provide gollum with details on where to find the plantuml.jar file?

I would just use a constant like `Gollum::UMLPlant` that a user could set in a `config.rb` file in gollum.

> How to provide gollum with the path for the java binary?

Same thing.

> The plantuml.jar file generates image files (.png). Where should these images be placed?

They have to be in a place readable by the webserver, of course. I would suggest a `tmp` or `uml` folder under gollum's asset folder. Probably that would require setting another constant, because I don't think `gollum-lib` knows where gollum's files are located. Good luck!

PR #135, opened 2015-01-16
Editable sections.

dometto:
PR to discuss pros/cons of this experimental implementation of https://github.com/gollum/gollum/issues/715

Another way of implementing this would be by parsing the document into sections client-sided, through the markup libraries already available as part of the ACE editor. However, I don't know if any them support extracting sections, and the problem remains of how to map id's generated by gollum's TOC filter to the raw markup.

### DIFF
--- a/gollum-lib.gemspec
+++ b/gollum-lib.gemspec
@@ -5,8 +5,8 @@ Gem::Specification.new do |s|
   s.required_ruby_version = '>= 1.9'
 
   s.name              = 'gollum-lib'
-  s.version           = '4.0.0'
-  s.date              = '2014-11-26'
+  s.version           = '4.0.1'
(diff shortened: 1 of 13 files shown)


dometto:
We're now considering moving to markdown.js for the livepreview (https://github.com/gollum/gollum/issues/826). markdown.js supports generating an [intermediate representation of the data](https://github.com/evilstreak/markdown-js#intermediate-representation), which we could possible use to split the sections up. We could look up the correct id's for each section via an ajax call. Possible pitfall is the fenced codeblocks: since markdown.js doesn't support these (yet https://github.com/evilstreak/markdown-js/issues/41), lines starting with `#` in a code block will probably be misinterpreted as section headings.

PR #136, opened 2015-01-26
Add support for PlantUML diagrams.

hsanson:
Second attempt at adding PlantUML diagrams to gollum-lib. My previous attempt generated the diagrams locally using the plantuml.jar program. This version now uses the PlantUML web service in a similar fashion as the WSD filter. Also add configuration capabilities to select the web service endpoint.

### DIFF
--- a/lib/gollum-lib/filter/plantuml.rb
+++ b/lib/gollum-lib/filter/plantuml.rb
@@ -1,0 +1,144 @@
+# ~*~ encoding: utf-8 ~*~
+require 'net/http'
+require 'uri'
+require 'open-uri'
+require 'zlib'
+
+# PlantUML Diagrams
+#
+# This filter replaces PlantUML blocks with HTML img tags. These img tags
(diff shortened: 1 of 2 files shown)


dometto:
I like this approach! Two things:

* It needs tests (using stubs/mocks to simulate the plantuml server)
* Can we give the user some feedback (instead of just a broken image tag) when the server is unresponsive or 404? Also useful for people who don't know that the PlantUML functionality exists. :)

Thanks for chipping in!

dometto:
For user feedback, even setting an `alt` attribute for the image may be enough - as long as there is an easy way way to see that that the missing image is a piece of UML not being rendered, rather than just any broken image.

dometto:
Ah, also, could you:
* Rebase everything into one commit after you've added the `alt` and tests?
* Then reopen this PR against the `rc` branch? (We reserve `master` for bugfixes/patch level changes only, we should update the Readme actually)

hsanson:
I will create a new pull request with the tests and user feedback.

PR #137, opened 2015-01-26
Only use the default email when name is empty

simonzack:
I've also added angle-bracket email stripping which is also related to the issue. Tests are modified so they pass.

See https://github.com/gollum/gollum/issues/954.

### DIFF
--- a/lib/gollum-lib/committer.rb
+++ b/lib/gollum-lib/committer.rb
@@ -53,8 +53,8 @@ module Gollum
     # Returns a Gollum::Git::Actor.
     def actor
       @actor ||= begin
-        @options[:name]  = @wiki.default_committer_name if @options[:name].to_s.empty?
-        @options[:email] = @wiki.default_committer_email if @options[:email].to_s.empty?
+        @options[:name]  = @wiki.default_committer_name if @options[:name].nil?
+        @options[:email] = @wiki.default_committer_email if @options[:email].nil?
         Gollum::Git::Actor.new(@options[:name], @options[:email])
       end
     end
--- a/lib/gollum-lib/wiki.rb
+++ b/lib/gollum-lib/wiki.rb
@@ -885,8 +885,9 @@ module Gollum
     #
     # Returns the String email address.
     def default_committer_email
-      @default_committer_email ||= \
-        @repo.config['user.email'] || self.class.default_committer_email
+      email = @repo.config['user.email']
+      email = email.delete('<>') if email
+      @default_committer_email ||= email || self.class.default_committer_email
     end
 
     # Gets the commit object for the given ref or sha.
--- a/test/test_committer.rb
+++ b/test/test_committer.rb
@@ -8,21 +8,34 @@ context "Wiki" do
 
   test "normalizes commit hash" do
     commit    = { :message => 'abc' }
-    name      = @wiki.repo.config['user.name'] || @wiki.default_committer_name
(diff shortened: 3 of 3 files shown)


lib/gollum-lib/committer.rb
```diff
@@ -53,8 +53,9 @@ def index
     # Returns a Gollum::Git::Actor.
     def actor
       @actor ||= begin
-        @options[:name]  = @wiki.default_committer_name if @options[:name].to_s.empty?
-        @options[:email] = @wiki.default_committer_email if @options[:email].to_s.empty?
+        name = @options[:name]
+        @options[:name]  = @wiki.default_committer_name if name.to_s.empty?
+        @options[:email] = @wiki.default_committer_email if name.to_s.empty?
```
dometto: Why not allow for the case of an empty email, but non-empty name?

lib/gollum-lib/wiki.rb
```diff
@@ -886,7 +886,7 @@ def default_committer_name
     # Returns the String email address.
     def default_committer_email
       @default_committer_email ||= \
-        @repo.config['user.email'] || self.class.default_committer_email
+        @repo.config['user.email'].delete('<>') || self.class.default_committer_email
```
dometto: This is causing the tests to fail in cases when `@repo.config['user.email]` is nil

lib/gollum-lib/committer.rb
simonzack: Hmm doesn't this allow it? If `name == 'tom'`, then none of them will be set to the default values.

dometto: Right, but if `name.to_s.empty`, then both will be set to the default values. So you can't have `name == 'tom'` and `email == ''`, I think?

lib/gollum-lib/wiki.rb
```diff
@@ -885,8 +885,9 @@ def default_committer_name
     #
     # Returns the String email address.
     def default_committer_email
-      @default_committer_email ||= \
-        @repo.config['user.email'] || self.class.default_committer_email
+      email = @repo.config['user.email']
+      email = email.delete('<>') if !email.nil?
```
dometto: That oughta fix it. I think just `if email` is more canonical, though. :)

lib/gollum-lib/committer.rb
simonzack: If `name == 'tom' isn't `name.to_s.empty?` false? I've tried to add this to the unit test from line 23.

dometto: Err, sorry. Yes, I was actually thinking of a case where the user explicitly sets `:name` to `''` and `:email` to `'whatever'`. Then it will set `:email` to the default value, which I think is weird.

The question is also whether setting an empty name should be allowed or not. If we want to allow it we should just change the lines to:
```ruby
@options[:name] = @wiki.default_committer_name if @options[:name].nil?
@options[:email] = @wiki.default_committer_email if @options[:email].nil?
```
Not sure what to think about setting empty names though.

simonzack: Would `@options[:email] = @wiki.default_committer_email` be more reasonable here then?

simonzack: Yeah I think that's more natural, I'll change this.

dometto:
Many thanks!

simonzack:
My pleasure :)

PR #138, opened 2015-01-27
Add support for PlantUML diagrams.

hsanson:
Third attempt to add [PlantUML]( http://plantuml.sourceforge.net) diagram support to Gollum-lib:

 - Uses external service instead of java binaries.
 - Checks for external service availability and shows an error if not available.
 - Add tests using a nice (but inaccurate) Lotr ring bearers diagram.


dometto:
:+1: Many thanks! Would be great if you could write a short blurp about this on the wiki (although there's no rush since this is in `rc` for now).

PR #139, opened 2015-01-27
improves the `AllPages` macro by adding links & paths

simonzack:
Screenshot:

sceenshot


dometto:
Nice. I think there's basically no difference between the AllPages Macro and the GlobalTOC Macro now, except the latter renders the list in the TOC div? If so we should probably make the GlobalTOC a subclass of this Macro for massive refactoring.

simonzack:
Ah I wasn't aware of the global toc macro. Your implementation looks better imo so I'll close this.

PR #144, opened 2015-04-05
JRuby support

dometto:


test/test_wiki.rb
```diff
@@ -539,7 +539,6 @@
 context "page_file_dir option" do
   setup do
     @path          = cloned_testpath('examples/page_file_dir')
-    @repo          = Gollum::Git::Repo.init(@path)
```
dometto: `@repo` not being used in this context, so this line is unnecessary.

dometto:
\0/

@bartkamphorst, @sunny: what say you?

Would it be a good idea to move https://github.com/repotag/gollum-lib_rjgit_adapter to the gollum org (and add travis tests, coverage, etc.)?

test/test_wiki.rb
bartkamphorst: Fair enough.

bartkamphorst:
Exciting to see all the lights green! :tada: Nice work with the gemspec. I'm in favor of merging.

Adding Travis, coverage, etc. is also a good idea at this point, moving the gem can be decided at a later time I think.

sunny:
Good job there, also like the way the gemspec is gemerated!

Shouldn't `lib/version.rb` be under `lib/gollum-lib/`?

gemspec.rb
```diff
@@ -0,0 +1,97 @@
+def specification(version, default_adapter, platform = nil)
+  Proc.new do |s|
+    s.specification_version = 2 if s.respond_to? :specification_version=
+    s.required_rubygems_version = Gem::Requirement.new('>= 0') if s.respond_to? :required_rubygems_version=
+    s.rubygems_version = '0.0.1'
```
bartkamphorst: Should this really be 0.0.1?

dometto:
Good point @sunny :) Are you also ok with using kramdown als the default markdown gem for gollum ( https://github.com/gollum/gollum/pull/994)? I was thinking of bumping gollum to 4.0 after that.

sunny:
Yup, sounds good to me.

PR #146, opened 2015-04-09
Update file view tests

dometto:
The nokogiri sanitization previously used to compare the output removed the "checked" attribute of the checkbox fields on MRI, but not on Jruby, causing tests to fail. Luckily we don't need the sanitization since we now have `assert_html_equal`.


bartkamphorst:
:+1:

PR #148, opened 2015-04-23
Implement Page#section_data.

dometto:


lib/gollum-lib/filter/sourcemap.rb
```diff
@@ -0,0 +1,24 @@
...
+# Source Maps
+#
+# Find lines corresponding to h1 headers for the present markup and insert markers indicating the line number in the raw source.
+
+class Gollum::Filter::SourceMap < Gollum::Filter
+  def extract(data)
+  	regex = Gollum::Markup.formats[@markup.format][:h1]
+    return data if regex.nil?
+    i = 0
+    data.lines.map do |line|
```
dometto: Ruby's `gsub' with a block unfortunately doesn't have a line index variable, so we need to split the data, loop over it to get the index and generate the markers, and then join it again. :-(

lib/gollum-lib/page.rb
```diff
@@ -242,6 +242,45 @@ def toc_data()
       markup_class.toc
     end
 
+    # Public: Returns the start and end line numbers of the nth section of the page as an array.
+    #
+    # section - Fixnum index indicating which section to return (base 1)
+    #
+    # Returns an Array of the start and end line numbers for the section.
+    def section_map(section)
```
dometto: I guess this should be a private method.

lib/gollum-lib/filter/sourcemap.rb
bartkamphorst: I think you should be able to chain `gsub` with `with_index` like so
```ruby
data.gsub(regex).with_index {| _, i | "#{$1}GOLLUMSRCMP#{i}GOLLUMSRCMP#{$2}" }
```

dometto: Interesting idea @bartkamphorst. Unfortunately the index will be the nth match, and not the line number. But we can work around this by matching *every* line, and only substituting if the line matches the further conditions for an h1, like so:
```ruby
1.9.3-p362 :071 > "a\n# b    f\na".gsub(/^(#\s+)?(\w+.+)?/).with_index do |_, i|
1.9.3-p362 :072 >     "#{$1}GOLLUMSRCMP#{i}GOLLUMSRCMP#{$2}" if $1 && $2
1.9.3-p362 :073?>   end
 => "a\n# GOLLUMSRCMP1GOLLUMSRCMPb    f\na" 
```
It's not very transparent but that may be worth it if it performs significantly better?

lib/gollum-lib/markups.rb
```diff
@@ -6,7 +6,7 @@ class Markup
         Kramdown::Document.new(content, :auto_ids => false, :input => "markdown").to_html
     }
 
-    register(:markdown,  "Markdown", :regexp => /md|mkdn?|mdown|markdown/)
+    register(:markdown,  "Markdown", :regexp => /md|mkdn?|mdown|markdown/, :h1 => /^(#\s+)(\w+.*)/)
```
SkyCrawl: @dometto 
I see where you're heading with matching two groups (whitespaces and content in this case) and substituting like this `#{$1}GOLLUMSRCMP#{i}GOLLUMSRCMP#{$2}`. **Smart!** However, I think there's a slightly better solution because we're only interested in `real-section-number => source-line-or-direct-offset` mappings. The resulting substituted raw data will not be presented and thus we're free to completely substitute the original header's content, as long as the new content is alphanumeric!

So, this should be a simpler (and a bit more markup-independent) solution:
* Matching - `if line =~ /^#(.*)$/`
* Substitution - `line[$1] = "GOLLUMSRCMP#{i}GOLLUMSRCMP"`

Since we're completely substituting the original content, `GOLLUMSRCMP#{i}GOLLUMSRCMP` might even turn to `#{i}`!

**NOTES:**
* Is this possible and reliable for all markups? I think header syntax for some markups is multi-line by definition...
* Maybe we should substitute the headers for direct character offsets (start position of the header syntax)? After all, that is what we want and what's best to use when substituting the section for a new version, later on.

lib/gollum-lib/filter/sourcemap.rb
SkyCrawl: Indeed. I don't think we can use MRI to directly "one-line" a solution that maps sections to a source line. On the other hand, it may not even be the best solution - see below for the note about offsets.

dometto: Yes, my workaround for @bartkamphorst's `#find_with_index` idea won't work when we need to map multi-line headers. :( Hopefully looping over `data.split` won't be too slow.

lib/gollum-lib/markups.rb
dometto: > is matching and mapping lines possible and reliable for all markups? I think header syntax for some markups could be multi-line...

I think multiline matches are doable. It's a bit of a hassle to contruct the right regexes, but we're doing currently doing something similar for code blocks.

> Maybe we should substitute the header content for direct character offset (start position of the header syntax) instead of line number?

By offset you mean [start-line number, end-line number]: good idea!

> `line.gsub(/^#{1}\K\s*.+$/, "GOLLUMSRCMP#{i}GOLLUMSRCMP")`

Thanks for the `\K` hint. This will obviously have to be a bit different if we're also going to need multiline matches, but I'll give it a shot!

SkyCrawl: > By offset you mean [start-line number, end-line number]: good idea!

By `character offset`, I meant the total number of characters before that character (from the start of the source text). Line numbers alone are more complicated to use (see your implementation of `section-data`).

If I had a single class named `EditSectionPageHandle` persisted in the user's session, the solution could be as simple as this:

**tokenize page source (private):**
```ruby
def tokenize(page_source)
    regex_sub_header_content = Gollum::Markup.formats[@markup.format][:h1]
    return data if regex_sub_header_content.nil?

    # this one HAS TO RELATE TO THE PAGE SOURCE, NOT THE TOKENIZED VERSION
    current_section_start_offset = 0

    # tokenize
    page_source.lines.map do |line|
        # replace section content for the token and return it:
        # NOTE: returns "line" if regex didn't match so no problem!
        line.gsub(regex_sub_header_content, "#{current_section_start_offset}")
        
        # before the next iteration:
        # NOTE: "line" contains the whitespaces so no problem!
        current_section_start_offset += line.length
    end.join
end
```

**get section (public):**
```ruby
def get_section(page_source, section_number)
    tokenized = tokenize(page_source)
    
    # TODO: render etc.
    # pseudocode:
    real_sections = nokogiri_doc.get_all_sections   # return all H1s by a css selector, for instance
    
    # section_number is 1-based, so:
    @start_offset = extract_token(real_sections[section_number - 1])
    @end_offset = extract_token(real_sections[section_number])
    if @end_offset.nil?
        # section_number is the last section of the page, in which case:
        @end_offset = page_source.length - 1
    else
        # a little adjustment for the below to work correctly
        @end_offset -= 1
    end
    
    # and return:
    page_source[@start_offset..@end_offset]
end
```

**get edited page (public):**
```ruby
def edited_page(page_source, new_section)
    raise RuntimeError, "The #get_section method has to be invoked first to determine section offsets." if @start_offset.nil? || @end_offset.nil?
    page_source[@start_offset..@end_offset] = new_section
end
```

I think this way, it's much clearer and more concise. And a little faster :).

dometto: Ah, that makes clear what you're after. I think something like `Page#edited_page` doesn't really fit gollum, since it doesn't play well with git -- it assumes we can edit the `Page`'s content, then save it to the repo, but there really isn't anything like a persisting, unsaved-but-modified page due to git's architecture (different content = different blob). That makes for less than ideally OO code, but that's the price of having git as a backend. :)

So since a `Page`'s content is basically static, I think there needs to be a method that returns the a page's content, split into `before`, `wanted section`, and `after`. That array can then be used in the frontend to display only the `wanted section`, joining it again with `before` and `after` when the form is submitted.

Nevertheless, interesting setup, and this:
```ruby
    @start_offset = extract_token(real_sections[section_number - 1])
    @end_offset = extract_token(real_sections[section_number])
```
Achieves what I was doing in a much cleaner way. :+1: I'll play around a bit and try to integrate as much of @SkyCrawl's and @bartkamphorst's suggestions.

SkyCrawl: > I think something like Page#edited_page doesn't really fit gollum, since it doesn't play well with git -- it assumes we can edit the Page's content, then save it to the repo, but there really isn't anything like a persisting, unsaved-but-modified page due to git's architecture (different content = different blob). That makes for less than ideally OO code, but that's the price of having git as a backend. :)

I'm sorry but I don't understand what you tried to say with this comment :p.
It seems I confused you again, sorry... should've made up a complete example :(. Let me clarify:

I think it would be quite troublesome to implement the "edit section" feature for irb processes. As such, I thought this feature would be frontend-exclusive (correct me if I'm wrong). If that's the case, suppose a class (frontend) like this:

```ruby
module Gollum
    class EditSectionPageHandle
        def initialize(path)
            # version is supposed to be the latest because... what's the point of editing a section of a historic version of a page?
            @page_source = ...      # use gollum-lib to read the page
            ...
        end
    end
    
    def get_section(section_number)
        # as defined above, except that this bit:
        #
        # ... render etc.
        # real_sections = nokogiri_doc.get_all_sections   # return all H1s by a css selector, for instance
        # 
        # should be defined in gollum-lib and called from the frontend (a.k.a. here)
    end
    
    def save_edited_page(new_section_content)
        wiki = ...
        edited_page = edited_page(new_section_content)     # as defined above
        wiki.update_page(page, name, format, editeddata, commit = {})
    end
    
    private
    
    def tokenize
        # as defined above
    end
    
    def edited_page(new_section)
        # as defined above
    end
```

**My original point was: define a frontend class to provide the "edit section" feature :).**

This frontend class clearly uses backend facilities to provide the "edit section" feature to the frontend. There are no "git issues", except for locking the page against other edits, which we have to solve anyway because otherwise, either one of these hinders Gollum's efforts:
* other edits cause the line numbers, offsets or even the `before, section, after` strings not to be reliable,
* saving the edited section causes the other edits to be ignored.

BTW, as far as I know, Gollum fails to protect users against this already with its editing features. A simple solution would be to make these features accessible only for Gollum instances bound to `localhost` (personal Wikis)...

Now, if instances of this class were stored in Sinatra's (or Rack's, not sure atm) user sessions, the process of updating the page (from frontend's point of view) is simple (pseudocode):

```ruby
map '/save-edited-page/...' do
    handler = Sinatra.get_session_for_request.get_page_handle(page_path)
    handler.save_edited_page(...)
end
```

SkyCrawl: > So since a Page's content is basically static, I think there needs to be a method that returns the a page's content, split into before, wanted section, and after. That array can then be used in the frontend to display only the wanted section, joining it again with before and after when the form is submitted.

Yes, that's another possibility :).

SkyCrawl: @dometto 
In case you'd be interested, I played around with the section regexes for all markups. Most of them are easy but some are downright problematic. You can see the result of my playing around in the section starting from [here](https://github.com/SkyCrawl/gollum-lib/blob/css-to-sass/lib/gollum-lib/configuration.rb#L98).

dometto: Thanks @SkyCrawl. I say we only activate the editable sections capability for markups that have an h1 regex defined. So if we can't come up with a good regex for org and rst, that's too bad for them. :)

SkyCrawl: Idd @dometto , I had the same thoughts :).

PR #161, opened 2015-06-12
Move search tests to right context, add tests to make sure search res…

dometto:
…ults do not contain false positives. See https://github.com/gollum/rugged_adapter/issues/7


dometto:
The rjgit adapter is not passing the new test that checks for false positives.

PR #162, opened 2015-06-26
fixed BlobEntry.normalize_dir()

myukselen:
On windows, expand_path may add "C:/" to the output. Added line strips
off this case.

Test suite is not running on windows.

### DIFF
--- a/lib/gollum-lib/blob_entry.rb
+++ b/lib/gollum-lib/blob_entry.rb
@@ -87,6 +87,7 @@ module Gollum
       return '' if dir =~ /^.:\/$/
       if dir
         dir = ::File.expand_path(dir, '/')
+        dir = dir[2..-1] if dir =~ /^[a-zA-Z]:\// # expand_path may add d:/ on windows
         dir = '' if dir == '/'
       end
       dir


dometto:
:+1: Thanks! Can you maybe let us know about your experiences running gollum on windows?

myukselen:
Thanks for the utility, I really loved ability for serving wiki pages inside my repositories.

On windows, I have installed jruby 1.7.20 straight ahead. Then installed gollum easily.
I had to debug with RubyMine until I figured out the problem.

I have installed latest TortoiseGit and Git-1.9.5-preview20150319.exe
I had to build gollum-lib inside Git-bash because of the "mv" command not found error. cmd.exe sucks.
Then I could install the fixed java version of the gem.
I start gollum inside the repo by standard cmd.exe. Now links and everything works fine for me.

I could not run the tests though. That is why no test cases added for the fix. :/

techwiz24:
@myukselen What was the error when trying to run tests? On my arch box tests won't run by default either because the 'test-unit' gem is missing by default, you can try adding the following to the end of the `Gemfile`:

```ruby
gem "test-unit"
```

myukselen:
I guess that helped. This is the output of the "bundle exec rake test":

Finished in 72.532 seconds.
--------------------------------------------------------------------------------

267 tests, 450 assertions, 16 failures, 39 errors, 0 pendings, 0 omissions, 0 no
tifications
83.1461% passed
--------------------------------------------------------------------------------

3.68 tests/s, 6.20 assertions/s
rake aborted!
SignalException: 1
D:\dev\ruby\jruby-1.7.20\bin\rake:23:in `(root)'
Tasks: TOP => test
(See full trace by running task with --trace)

myukselen:
The test aimed at normalize_dir was already failing but the change committed fixed the test.

$ bundle exec jruby test/test_page.rb -n test_normalize_dir
io/console not supported; tty will not be manipulated
io/console not supported; tty will not be manipulated
Loaded suite test/test_page
Started
.

Finished in 0.09 seconds.
--------------------------------------------------------------------------------

1 tests, 6 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifica
tions
100% passed
--------------------------------------------------------------------------------

11.11 tests/s, 66.67 assertions/s

bartkamphorst:
Thanks @myukselen ! Very interesting. In the 16 failures and 39 errors, are there any patterns discernible?

myukselen:
It seems most of them are file accessing and modification related issues. Differences between Windows and *nix style file handling. Some of them are burried into rjgit, some are surfaced in fileutils.rb

Also I see utf8 handling issues but these can be related to my locale settings.

If you are intereseted and can suggest a method, I will upload 190k test output.

dometto:
I suposse lots of things like hardcoded slashes instead of using File::SEPARATOR :) (if I'm correct in assuming Windows still uses backslashes?)

To upload the test output, why not use a github gist? Thanks for the assistance!

myukselen:
You are welcome :)

I have uploaded as gist, here is the link:
https://gist.github.com/myukselen/553d630fadf8da19f652#file-gollum-lib-test-output-on-jruby1-7-windows

PR #163, opened 2015-07-02
Add macro to display page tree with links.

flozzone:
Pages and Directories are displayed as nested unordered lists, where pages
link to their page and directories link to theirs index page.

Signed-off-by: flozzone <flozzone@gmail.com>


dometto:
Sorry for the slow reaction, but thanks for the PR! :+1: I like this, but I'm wondering to what extent we still need a separate `GlobalTOC()` (which is now in the `rc` branch) and `AllPages()` macro if we're going to merge this. Could the two macro's somehow be merged?

dometto:
ping @flozzone.

flozzone:
Hi, I'm currently on holidays and was afk ;). 

In my opinion GlobalTOC() completely replaces AllPages(), because it does the same, but displays the path to the file and a link to it. The Navigation() macro does nearly the same as GlobalTOC() but keeps the directory hierarchy as nested unordered lists, instead of one flat list. Another difference is that Navigation() also displays the directory as a link (if there is a index_page in it), so that the user gets pointed to the directory index page.

Currently GlobalTOC() doesn't work well inside the sidebar, because it uses relative links, which gives a wrong link when trying to visit a page in another subtree than the current one. Making the links absolute would fix this problem.

Generally Navigation() replaces GlobalTOC(). AllPages() was useless for me, because it doesn't provide any functionality, but I used it as a starting point for the Navigation() macro.

PS: The name "GlobalTOC" is somehow a bit misleading in my opinion, because it lists pages and not the headers inside the pages. Maybe "FlatNavigation" would be more appropriate.

flozzone:
I've rebased on master and made the macro more compatible (by using @wiki.index_page) and added a test directory structure.

dometto:
Appreciate you chipping in on holiday. :)

I'm allright with Navigation() replacing GlobalTOC(). However, we just 
released the first version of gollum-lib containing GlobalTOC(), so 
getting rid of it so soon is maybe a bit confusing to the end user. On 
the one hand, I get your point about GlobalTOC being slightly confusing. 
On the other, to keep some kind of backwards compatibility, it would be 
nice to just rename Navigation() to GlobalTOC(), and get rid of the 
current AllPages and GlobalTOC altogether. Do you feel very strongly 
about the name? Another option is to have `Macro::GlobalTOC` subclass 
`Macro::Navigation` and not override anything, so that both GlobalTOC() 
and Navigation() would work. But then at some point we would have to 
break backward compatibility anyway.

flozzone:
I don't feel strongly at all about the name, it's better to keep compatibility. The idea of having a navigation helper class in the macro scope is nice. The helper functions to build up a tree of pages using hashes, could be reused by other macro editors, or is there already something in place that could replace this logic?

dometto:
Sounds good. If you like you can change the name of your macro to GlobalTOC then, and remove AllPages and the current GlobalTOC. Also please squash your commits when done. :)

There's no rush of course, so enjoy your holiday!

Regarding the tree hash, Wiki#tree_map might be helpful?

dometto:
> Sounds good. If you like you can change the name of your macro to GlobalTOC then, and remove AllPages and the current GlobalTOC. Also please squash your commits when done. :)

@flozzone, will you have time to make these final changes to the PR?

flozzone:
Regarding Wiki#tree_map: I see Wiki#tree_map_for which returns me a flat list of blob entries, but its easier to build up a recursive list, this is why I'm building a tree of maps (pages as leafs and folders as nodes) first and then visit every node and leaf recursively. I have some concerns about the efficiency of the algorithm, because this tree is built up every time a page is requested.

Isn't there any function that returns a real recursive tree of git::blobs and git::trees, that I can visit recursively? Another solution would be to build the tree on the client side via javascript, which then could handle collapsing and expanding different nodes.

dometto:
> I see Wiki#tree_map_for which returns me a flat list of blob entries

Ah, I was under the impression that would return a hash tree of blobs and trees. I can get a sortof recursive list by doing the following (after running `bundle exec gollum --irb` from the gollum repo):
```ruby
 a = Gollum::GitAccess.new(wiki.path, nil, nil)
a.repo.lstree(wiki.ref, {:recursive => true})
```
Excerpt from output:
```
[{:mode=>"040000", :type=>"tree", :sha=>"08316088e67dc686d1390fbf83177c51ca341143", :path=>"lib"}, {:mode=>"100644", :type=>"blob", :sha=>"55b81ca64cb2ed732a2f58ed4f204a5fade1db1f", :path=>"lib/gollum.rb", :size=>935}, {:mode=>"040000", :type=>"tree", :sha=>"16203cfde01c44758196438458d02f495060bc37", :path=>"lib/gollum"}, {:mode=>"100644", :type=>"blob", :sha=>"12e47541a1196d06bc6cf1ec4c3421358fab1568", :path=>"lib/gollum/app.rb", :size=>17917}]
```
So it's an array of hashes representing blobs and trees, sorted by filepath:
```ruby
 a.repo.lstree(wiki.ref, {:recursive => true}).each {|x| p x[:path]}
".gitattributes"
".gitignore"
".travis.yml"
"Gemfile"
"HISTORY.md"
"LICENSE"
"README.md"
"Rakefile"
"bin"
"bin/gollum"
"config.rb"
"config.ru"
"contrib"
"contrib/openrc"
"contrib/openrc/conf.d"
"contrib/openrc/conf.d/gollum"
"contrib/openrc/init.d"
"contrib/openrc/init.d/gollum"
"contrib/systemd"
"contrib/systemd/gollum@.service"
"contrib/sysv-debian"
"contrib/sysv-debian/init.d"
"contrib/sysv-debian/init.d/gollum"
"gollum.gemspec"
"lib"
"lib/gollum.rb"
"lib/gollum"
"lib/gollum/app.rb"
```

If that's helpful to you, we can add a method like this to `GitAccess` and `Wiki`, so that the result will also be cached.

dometto:
Another possibility is to interact directly with the underlying Git abstraction layer, for example:
```
a.repo.index.current_tree.blobs
 [#<Gollum::Git::Blob:0x007f96ec5256e0 @blob=#<Grit::Blob "09ab669cc721e976ac25a7915286c2b5eb12b6da">>, etc.]
```
However, there is not yet a corresponding method to get all the trees in current_tree. We would have to add that to the adapter specifications, and then implement it for each adapter.

flozzone:
This is exactly what I was looking for. I will implement it with this method.

rgroux:
hi 
Actualy i had already code a macro that create a Sub Global TOC. By default it will print all pages, but you can also print only a sub folder.
This macro can use (if present) the meta inside the page like title and description  instead of page name and page path.

Something like on the snapshot.

snapshot5

The code is a little ugly, so i will refactor and make it better. I think an optional parameter for  name+path/title+description should be better.

If you are interested tell me, i will create a PR for this, i also have a macro for jumly uml, i can  create a PR for that too.

/Richard

calid:
Hello, is there any ETA for when this PR will be merged and go live as the new GlobalTOC/AllPages?  Because of the relative links issue the current GlobalTOC isn't all that useful.

I'm happy to chip in if there's any help needed to get this out the door.

Cheers!

calid:
After playing with @flozzone's Navigation macro a little further, I think @rgroux's idea of optionally only printing navigation links for the current subtree would be really nice to incorporate.

rgroux:
@calid

done in   #182 
without the meta, there are a bug, so i will work on it ASAP, but you can actualy with this code get the TOC with links from top lovel or sublevel or a regexp like

SubGlobalTOC("(Pages in this Wiki", "") == SubGlobalTOC("(Pages in this Wiki", "") == GlobalTOC("(Pages in this Wiki")

SubGlobalTOC("(Pages in this Wiki", "^FirstLevel") => TOC from sub folder FirstLevel
SubGlobalTOC("(Pages in this Wiki", "Test") => all pages with Test in the name

hope it will be accepted, and used and hack/extend in a near future

calid:
@rgroux thanks, I'll try it out.  Just to clarify, by default will SubGlobalTOC use the current page as the root of the TOC tree, or do you always have to explicitly give the desired sublevel root?  i.e. given the following structure:

```
foo
├── bar
│   ├── Bar.md
│   └── baz
│       └── Baz.md
└── Foo.md
```

If I use `SubGlobalTOC("Pages in this Tree")` in `Bar.md`, will I get a TOC of:

> **Pages in this Tree:**
> Bar
> baz/Baz

rgroux:
not exactly
SubGlobalTOC("Pages in this Tree","bar")
foo/bar
foo/baz/baz

I keep the full path, hackable with a boolean and a 3rd parameter. In my case i need the full path, but it can be useful to only show relative path.

i think we must wait the PR is accepted and after hack it and add some features.
I use this macro to show all pages relative to some features like this in tree
.
├── Dev
│   ├── release
│   │   ├── Action.md
│   │   └── Plan.md
│   └── test
│       ├── Action.md
│       └── Plan.md
├── Pre
│   ├── release
│   │   ├── Action.md
│   │   └── Plan.md
│   └── test
│       ├── Action.md
│       └── Plan.md
├── Prod
│   ├── release
│   │   ├── Action.md
│   │   └── Plan.md
│   └── test
│       ├── Action.md
│       └── Plan.md

```SubGlobalTOC(Pages in this Wiki,"Plan")```
>Pre/release/Plan.md
>Pre/test/Plan.md
>Dev/test/Plan.md
>Dev/release/Plan.md
>Prod/release/Plan.md
>Prod/test/Plan.md

and 
```SubGlobalTOC(Pages in this Wiki,"Dev")```
>Dev/test/Action.md
>Dev/test/Plan.md
>Dev/release/Action.md
>Dev/release/Plan.md

calid:
One of the reasons I like the implicit root is that it will handle renames and reorgs transparently... if I have to explicitly define the root or match pattern now I have something else to update when I rename pages or shuffle my wiki around...

Maybe make the default root-at-current-page, and just override that if a parameter is passed?  I think this would be really nice (and intuitive) default behavior

rgroux:
indeed, 

when i wrote this macro i was in a lazy mode :p

flozzone:
Thanks for taking over this feature.

dometto:
Got a little bit lost here. I guess the idea is that this PR has been superseded by #182? Then we should close this and move the discussion there. @flozzone: please let me know if you see good reasons to keep this open instead of or in addition to #182.

flozzone:
I haven't tested #182 yet, but it seems that @rgroux has same intentions as I have. I will close this pull request and follow the discussion in #182.

PR #164, opened 2015-07-21
Add feature julmy UML diagram

rgroux:
In the pull request #134 jumly is presented as a better solution for
UML than plantuml.

I reuse the work of wds and adapt it for [Jumly](http://jumly.tmtk.net/) as it show in the jumly [REST API](https://github.com/tmtk75/jumly/wiki/REST-API)

### DIFF
--- a/lib/gollum-lib/filter/jumly.rb
+++ b/lib/gollum-lib/filter/jumly.rb
@@ -1,0 +1,48 @@
+# ~*~ encoding: utf-8 ~*~
+require 'net/http'
+require 'uri'
+require 'open-uri'
+
+# Jumly UML Diagrams
+#
+# Render an inline jumly uml diagram by sending the jumly code through the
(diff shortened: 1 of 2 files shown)


dometto:
Unfortunately, it seems like Jumly, like WebSequenceDiagrams, is dependent on an external platform. That was the reason we want to move to PlantUML. So I think we won't merge this. However, if you want to use Jumly, you can of course convert this implementation to a custom macro.

rgroux:
@dometto 
ok, but I am a little confuse, because you say in another PR jumly can be a better choice (than plantuml).
But anyway i understand you don't want to depend on remote service.

The Licence of [jumly](http://jumly.tmtk.net/) is [MIT](http://opensource.org/licenses/MIT), can we build and add the `jumly.min.js` file and route queries instead of  `http://jumly.tmtk.net/api/diagrams?data=`.


/Richard

PR #165, opened 2015-08-09
Version 4.1

dometto:
`rc` contains a lot of bug fixes and some nice features, like plantuml support. We were hoping to get CSS updates in there, but I don't see this happening in the near future. So let's get this stuff out there. :-)


bartkamphorst:
:+1: Thanks @dometto !

PR #168, opened 2015-08-22
Fix #154: Code block not correctly escaped when using pygments

### LINKED ISSUE #154: Code block not correctly escaped when using pygments
Putting 

    
    ```bash
    sudo patch -p1 < ~/auth.patch
    ```
    
in a markdown file causes the remainder of the document to be rendered inside of a `<pre>` tag. Removing the `<` character from the code block corrects the problem. This problem is not encountered if `pygments.rb` is not installed.

techwiz24:
For some reason code blocks used to default to simple `<pre>` tags when the specified language was nil or bash. This caused html tags in the code block to not get escaped since they were never passed
to pygments.

This fix defaults to the 'text' language if none is specified and allows pygments to render bash code blocks since it supports it.

### DIFF
--- a/lib/gollum-lib/filter/code.rb
+++ b/lib/gollum-lib/filter/code.rb
@@ -72,16 +72,12 @@ class Gollum::Filter::Code < Gollum::Filter
       encoding = @markup.encoding || 'utf-8'
 
       if defined? Pygments
-        # treat unknown and bash as standard pre tags
-        if !lang || lang.downcase == 'bash'
-          hl_code = "<pre>#{code}</pre>"
-        else
-          # Set the default lexer to 'text' to prevent #153
-          lexer = Pygments::Lexer[(lang)] || Pygments::Lexer['text']
+        # Set the default lexer to 'text' to prevent #153 and #154
+        lang = lang || 'text'
+        lexer = Pygments::Lexer[(lang)] || Pygments::Lexer['text']
 
-          # must set startinline to true for php to be highlighted without <?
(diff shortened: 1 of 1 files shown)


techwiz24:
screenshot from 2015-08-22 10-39-52

I'm going to try to write a test for this, which is why I have marked the PR as `[WIP]`

bartkamphorst:
Excellent. Looks good to me!

techwiz24:
@bartkamphorst Two quick things:

1. There doesn't seem to be an existing set of tests for the `code` filter. Do you still want one written for this PR?
2. If you want one written, we will need to require `pygments.rb` as a test dependency. How do you want to proceed with that?

dometto:
Great work! Tests are always nice. If you have time to come up with some rudimentary ones, that would be great, but otherwise we can just merge this as a bugfix PR.

Adding pygments as a development dependency would break jruby support, so maybe we could just stub `Pygments::Lexer`? Testing whether Pygments is being called should be enough.

techwiz24:
@dometto Of course! I didn't even think of a mock / stub!

I'm going to go ahead and say lets just merge this now. I need some time to get the basics of tests / mocks in ruby (I'm still not terribly familiar with the language itself). Perhaps myself or someone else can make a PR for tests at a later time? 

:white_check_mark: Ready for merging at your leisure, tests will come in another PR.

dometto:
@techwiz24, sorry for leaving this open so long!

PR #174, opened 2015-10-03
Fixed changing wiki format and renaming with page_file_dir

mpnowacki:


mpnowacki:
Okay, so after doing some more thinking I came to a conclusion, that this problem can be solved in a better way. The problem lies in calling `add_to_index`. It's called three times in `gollum-lib/wiki.rb`:
* `def write_page` - works fine, didn't touch it
* `def update_page` - gets too many `page_file_dir`s
* `def rename` - same as above
And one time in `gollum/app.rb`, in `post '/uploadFile'`, where, in case of `--allow_uploads=page` the `dir` argument is deprived of `page_file_dir` using a regexp.
That makes five calls to `add_to_index` (I'm counting `uploadFile` twice, because of the modes it's called in). Three of them need slicing `page_file_dir` out. It seems to me it will be nicer to make `add_to_index` not join the `page_file_dir` part and add explicit joining of it in `write_page` and `/uploadFile`. I think I'll try and do that.

mpnowacki:
okay, i've fixed the problem in a nicer way, there are some chnges in gollum as well. They can be found here: https://github.com/gollum/gollum/pull/1077

dometto:
Thanks for the work! Think your new strategy is better indeed. I'll try to inspect it more closely this afternoon.

mpnowacki:
Another possible strategy, would be to make the client `page_file_dir` agnostic and automatically prepend it to every path received in every service. Thus could be a bit limiting, because it wouldn't be possible to view files not in `page_file_dir`, but I'm not sure if such actions are safe anyway. 
In this approach, the address of the home page (the one that the browser uses) would always be `host/Home`, regardless of `page_file_dir`. But that's a major change, and this bug probably needs to be solved quickly.

lib/gollum-lib/committer.rb
```diff
@@ -94,7 +94,7 @@ def add_to_index(dir, name, format, data, allow_same_ext = false)
 
       dir  = '/' if dir.strip.empty?
 
-      fullpath = ::File.join(*[@wiki.page_file_dir, dir, path].compact)
+      fullpath = ::File.join(*[dir, path].compact)
```
dometto: I think we don't need the `compact` anymore, since both `dir` and `path` shouldn't be nil at this point. So just `::File.join(dir, path)` should suffice.

dometto:
> Another possible strategy would be to make the client page_file_dir agnostic and automatically prepend it to every path received in every service. Thus could be a bit limiting, because it wouldn't be possible to view files not in page_file_dir, but I'm not sure if such actions are safe anyway.

Yes, I would be in favor of that. It seems like what people intuitively expect from the `--page-file-dir` option anyway. See https://github.com/gollum/gollum/issues/1052

lib/gollum-lib/committer.rb
mpnowacki: I agree

PR #176, opened 2015-10-16
fix for toc list levels

bpieck:
I recognized, when I overwrite your toc from using ul to ol, that the list levels are kind of a little messed up.
I fixed that for me on ol - but I think perhaps you want to have it on the ul, too. With this it's easy to perhaps add an option, to use ol instead of ul. Only line 89 would have to be changed for that.


dometto:
Sorry for the slow reply and thanks for the PR @bpieck. I'll try to inspect this later today!

dometto:
Again sorry for the slowness. Thanks @bpieck!

PR #181, opened 2015-11-10
Add Page#last_version, as described in gollum/gollum#1087.

jhominal:
* Define `Page#last_version`;
 * Add some test assertions related to `#version` and `#last_version`;

Feel free to comment on it and ask for amendments.

### DIFF
--- a/lib/gollum-lib/page.rb
+++ b/lib/gollum-lib/page.rb
@@ -281,6 +281,14 @@ module Gollum
       @wiki.repo.git.versions_for_path(@path, @wiki.ref, log_pagination_options(options))
     end
 
+    # Public: The last version that has touched the Page. Can be nil.
+    #
+    # Returns Gollum::Git::Commit, or nil.
+    def last_version
+      return @last_version if defined? @last_version
+      @last_version = @wiki.repo.git.versions_for_path(@path, @wiki.ref, {:max_count => 1}).first
+    end
+
     # Public: The first 7 characters of the current version.
     #
     # Returns the first 7 characters of the current version.
--- a/test/test_page.rb
+++ b/test/test_page.rb
@@ -24,6 +24,10 @@ context "Page" do
     assert_equal 'Bilbo-Baggins.md', page.path
     assert_equal :markdown, page.format
     assert_equal @wiki.repo.commits.first.id, page.version.id
+
+    assert_not_nil page.last_version
+    assert_equal page.versions.first.id, page.last_version.id
+    assert page.last_version.stats.files.map{|file_stats| file_stats[0]}.include? page.path
   end
 
   test "get existing page case insensitive" do
--- a/test/test_page.rb
+++ b/test/test_page.rb
@@ -27,7 +27,7 @@ context "Page" do
 
     assert_not_nil page.last_version
     assert_equal page.versions.first.id, page.last_version.id
-    assert page.last_version.stats.files.map{|file_stats| file_stats[0]}.include? page.path
+    assert page.last_version.stats.files.map{|file| file_path = file.first}.include? page.path
   end
 
   test "get existing page case insensitive" do
--- a/test/test_wiki.rb
+++ b/test/test_wiki.rb
@@ -130,6 +130,10 @@ context "Wiki page previewing" do
     assert_html_equal "<h1><a class=\"anchor\" id=\"bilbo\" href=\"#bilbo\"><i class=\"fa fa-link\"></i></a>Bilbo</h1>", page.formatted_data
     assert_equal "Test.md", page.filename
     assert_equal "Test", page.name
+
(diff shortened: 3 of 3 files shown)


lib/gollum-lib/page.rb
```diff
@@ -281,6 +281,14 @@ def versions(options = {})
       @wiki.repo.git.versions_for_path(@path, @wiki.ref, log_pagination_options(options))
     end
 
+    # Public: The last version that has touched the Page. Can be nil.
+    #
+    # Returns Gollum::Git::Commit, or nil.
+    def last_version
+      return @last_version if defined? @last_version
```
bartkamphorst: `defined?` will return the string `"instance-variable"` (which is truthy) even if `@last_version` is nil.
```
jruby-1.7.19 :001 > @last_version = nil
 => nil 
jruby-1.7.19 :002 > defined? @last_version
 => "instance-variable"
```
Is this the behavior you want? I would think `return @last_version if @last_version` is preferable, or not?

test/test_page.rb
```diff
@@ -24,6 +24,10 @@
     assert_equal 'Bilbo-Baggins.md', page.path
     assert_equal :markdown, page.format
     assert_equal @wiki.repo.commits.first.id, page.version.id
+
+    assert_not_nil page.last_version
+    assert_equal page.versions.first.id, page.last_version.id
+    assert page.last_version.stats.files.map{|file_stats| file_stats[0]}.include? page.path
```
bartkamphorst: Can we make more explicit what `file_stats` are and why we need to inspect (only) the first element?

lib/gollum-lib/page.rb
dometto: This is the intended behavior, because we want to cache a possible `nil` value for `@last_version` (it will be `nil` if the Page is a preview).

bartkamphorst: Ah, I see. Carry on!

test/test_page.rb
dometto: Maybe just make it a two-liner for more clarity then?
```ruby
modified_file_paths = page.last_version.stats.files.map {|file_stats| file_stats.first}
assert modified_file_paths.include? page.path
```

bartkamphorst: That helps, but I guess I'm just confused about the `map` on that first line (semantically, not syntactically).

dometto: It's because `#stats` return [this Struct](https://github.com/gollum/rugged_adapter/blob/master/lib/rugged_adapter/git_layer_rugged.rb#L145). The structure of the struct (pun intended) is not very transparent, which I guess is causing the confusion. Another legacy from `grit`...

lib/gollum-lib/page.rb
jhominal: `#last_version` will only be `nil` if the page does not exist on disk yet. While the utility of returning a non-`nil` value in a preview is not very high, that is the current behavior so I kept it.

test/test_page.rb
bartkamphorst: Thanks. Yes, the confusion was caused by the fact that files are represented as arrays with 4 elements. The following creates a little bit of overhead but might be more legible?
```ruby
files.map{ | file | path, _, _, _, = file; path } 
```
Or simply
```ruby
files.map{ | file | file_path = file.first } 
```

bartkamphorst: Or we could refactor `#build_stats` such that files are also structs:
```ruby
files << OpenStruct.new(path: patch.delta.new_file[:path].force_encoding("UTF-8"), deletions: new_deletions, additions: new_additions, changes: patch.changes)
```
Not sure how much other stuff that would break, but that way we could do `files.map{ | file | file.path }`.

dometto: That's definitely the pretties way to go, but it requires changes in the 
two other adapters and both gollum and gollum-lib. I'm all far that, but 
maybe that could be in a different set of PRs? I think your suggestion 
of `files.map{ | file | file_path = file.first}` is sufficiently explicit.

jhominal:
The `files.map` block has been modified according to your suggestions.

As for breaking usages of `stats.files`, as far as I can see, there are two usages of that property between gollum-lib and gollum:

 * In gollum, in [`views/latest_changes.rb`](https://github.com/gollum/gollum/blame/v4.0.1/lib/gollum/views/latest_changes.rb#L24-L30)

jhominal:
Hello,

I was just wondering where this PR was at - do you have any more comments?

dometto:
Sorry about the delay. As far as I'm concerned this is ready to merge. I noticed that the newest version of rugged supports it's own #stats method, so that probably counts in favour of rewriting the stats implementation. But as I said that can be in different PRs.

dometto:
Thanks again @jhominal

PR #182, opened 2015-11-19
feature(macro): add a sub global toc with links

rgroux:
Hi,

like i said in #163, and for @calid, i have "Sub Global TOC" in y config.ru.

I made this PR and migrate the macro into macro lib folder in the gollum-lib, there are also an embryonic test (a copy of GlobalTOC).

### DIFF
--- a/lib/gollum-lib/macro/navigation.rb
+++ b/lib/gollum-lib/macro/navigation.rb
@@ -1,7 +1,7 @@
 module Gollum
   class Macro
     class Navigation < Gollum::Macro
-      def render(title = "Navigate in the TOC", toc_root_path = File.dirname(@page.path), full_path = false)
+      def render(title = "Navigate in the TOC", toc_root_path = ::File.dirname(@page.path), full_path = false)
         if @wiki.pages.size > 0
           result = '<ul>' + @wiki.pages.map {|p|
             if p.url_path.start_with?(toc_root_path)
--- a/lib/gollum-lib/macro/navigation.rb
+++ b/lib/gollum-lib/macro/navigation.rb
@@ -1,11 +1,12 @@
 module Gollum
   class Macro
     class Navigation < Gollum::Macro
-      def render(title = "Navigate in the TOC", toc_root_path = "")
+      def render(title = "Navigate in the TOC", toc_root_path = File.dirname(@page.path), full_path = false)
         if @wiki.pages.size > 0
           result = '<ul>' + @wiki.pages.map {|p|
-            if p.url_path.match(toc_root_path)
-              "<li><a href=\"#{p.url_path}\">#{p.url_path_display}</a></li>"
+            if p.url_path.start_with?(toc_root_path)
+	      path_display = full_path ? p.url_path_display  : p.url_path_display.sub(toc_root_path,"").sub(/^\//,'')
+              "<li><a href=\"/#{p.url_path}\">#{path_display}</a></li>"
             end
 	  }.join + '</ul>'
         end
--- a/lib/gollum-lib/macro/navigation.rb
+++ b/lib/gollum-lib/macro/navigation.rb
@@ -3,12 +3,13 @@ module Gollum
     class Navigation < Gollum::Macro
       def render(title = "Navigate in the TOC", toc_root_path = ::File.dirname(@page.path), full_path = false)
         if @wiki.pages.size > 0
-          result = '<ul>' + @wiki.pages.map {|p|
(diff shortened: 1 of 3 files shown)


rgroux:
A short description of the PR:

This macro extend the GlobalTOC, with a second argument, if empty SubGlobalTOC and GlobalTOC are identical, but with  second argument, we can filter with match the page we want to expose in the TOC. it can be useful for TOC of subfolder or a TOC type of page.

SubGlobalTOC("(Pages in this Wiki", "") == SubGlobalTOC("(Pages in this Wiki", "") == GlobalTOC("(Pages in this Wiki")

SubGlobalTOC("(Pages in this Wiki", "^FirstLevel") => TOC from sub folder FirstLevel
SubGlobalTOC("(Pages in this Wiki", "Test") => all pages with Test in the name

hope it will be accepted, and used and hack/extend in a near future

dometto:
@calid in #163:
> Maybe make the default root-at-current-page, and just override that if a parameter is passed? I think this would be really nice (and intuitive) default behavior

I agree with that. Would it be too much work to implement this in this PR?

dometto:
Also I think the name `SubGlobalTOC` is a bit complicated and not very clear (since it's precisely not a global TOC). Maybe take up a suggestion by @flozzone and call it something like `Navigation()`?

rgroux:
@dometto 
:+1:  for Navigation(), great name

for root-at-current-page, i will do it ASAP, today or tomorow.

rgroux:
@dometto 

i think it's ok now

lib/gollum-lib/macro/navigation.rb
```diff
@@ -0,0 +1,17 @@
+module Gollum
+  class Macro
+    class Navigation < Gollum::Macro
+      def render(title = "Navigate in the TOC", toc_root_path = ::File.dirname(@page.path), full_path = false)
+        if @wiki.pages.size > 0
+          result = '<ul>' + @wiki.pages.map {|p|
```
bartkamphorst: As a stylistic suggestion, you might consider the following.
```ruby
    path_display = full_path ? page.url_path_display  : page.url_path_display.sub(toc_root_path,"").sub(/^\//,'')
    "<li><a href=\"/#{page.url_path}\">#{path_display}</a></li>"
  end
result = "<ul>#{list_items.join}</ul>"
```

rgroux:
@bartkamphorst 

yes, easy and more readable :)

dometto:
:+1: Thanks!

calid:
hmm so I just installed gollum-lib off of master, and unfortunately this is not working for me (TOC is always empty).  

It seems to be because `toc_root_path` always resolves to `.` but `page.url_path` resolves to the name of the page without a `./` prefix, e.g. `Home`.  

So this condition never passes and pages never get added to the TOC:

```ruby
    path_display = full_path ? page.url_path_display  : page.url_path_display.sub(toc_root_path,"").sub(/^\//,'')
    "<li><a href=\"/#{page.url_path}\">#{path_display}</a></li>"
end
```

To debug I added this else condition:
```ruby
else
    "<li>'#{toc_root_path}' != '#{page.url_path}'</li>"
end
```

which in my test wiki prints:
*    '.' != 'Home'
*    '.' != 'gollum/Gollum'
*     '.' != 'lists/ToDo'

Would it be too simplistic to just prefix `page.url_path` with `./` ?

calid:
Actually, yeah it would... looks like even in nested wiki dirs toc_root_path still just resolves to `.`

dometto:
Meh, apologies for the uncareful merge. Should have done some manual testing. @rgroux, any ideas for a fix?

calid:
I wonder if #163 was prematurely closed.  I've been using @flozzone's branch locally and it works well, just missing the 'root at current subtree by default' logic.

@rgroux @flozzone's implementation seems more robust and already works with respect to pretty printing the TOC links as a tree.  Instead of reimplementing that logic what about just adding the subtree rooting and metadata title features to his implementation?

dometto:
That would be my preference, too.

rgroux:
@calid, @dometto 

sorry, for the bug. no idea actually, and i will be in a rush next week, i wil ltry a to look at it. but no guaranties.

little ideas for toc_root_path =::File.dirname(@page.path)
  - the wiki_path in config.ru
  - the wiki_path in command line
  - or any path in the config.

hope it will be this and easy to check and fix it

polyzen:
Bump

PR #183, opened 2015-11-22
Allow setting max header level in TOC

ztlpn:
Hi! This PR implements gollum/gollum#650.
`levels` attr is added to the `[[_TOC_]]` tag. Its value determines the maximum header level to appear in TOC.

For this to work with subpages I had to re-parse generated TOC html which is inelegant but I guess it is minimally invasive.

Also some empty `<ul>` tags are left in HTML but I did not bother fixing it because they are not rendered in browser and anyway the problem should go away after #176 is merged.

### DIFF
--- a/lib/gollum-lib/filter/toc.rb
+++ b/lib/gollum-lib/filter/toc.rb
@@ -34,15 +34,11 @@ class Gollum::Filter::TOC < Gollum::Filter
 
     @markup.toc = toc_str
 
-    data.gsub(/\[\[_TOC_(.*?)\]\]/) do
+    data.gsub!(/\[\[_TOC_(.*?)\]\]/) do
       levels = nil
-
-      opts = Regexp.last_match[1].split('|')[1..-1] || []
-      opts.each do |attr|
-        parts = attr.split('=').map { |x| x.strip }
-        if parts[0] == 'levels'
-          levels = Integer(parts[1])
-        end
+      levels_match = Regexp.last_match[1].match /\|\s*levels\s*=\s*(\d+)/
+      if levels_match
+        levels = levels_match[1].to_i
       end
 
       if levels.nil? || toc_str.empty?
@@ -58,6 +54,8 @@ class Gollum::Filter::TOC < Gollum::Filter
         toc_clone.to_xml(@markup.to_xml_opts)
       end
     end
+
+    data
   end
 
   private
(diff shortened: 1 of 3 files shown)


dometto:
Sorry for the slow reply, but thanks a lot for this! @ztlpn! However, a new bug relating to the TOC has just been reported: https://github.com/gollum/gollum/issues/1096 Can you check if your new code still contains this bug? If not, would you mind seeing if it can be fixed easily? I'd like to release a new gollum-lib gem soon and it would be great to have a new, well-functioning TOC in there.

jplitza:
@dometto Actually, gollum/gollum#1096 isn't about the TOC at all, it is simply about the generation of links in the normal page body. I'd be surprised if it had anything to do with this PR.

dometto:
Thanks for the explanation @jplitza. Will merge this then, unless anyone objects? /cc @bartkamphorst

lib/gollum-lib/filter/toc.rb
```diff
@@ -24,15 +25,38 @@ def process(data)
...
-      
     end
- 
-    @markup.toc = @toc
-    data.gsub("[[_TOC_]]") do
-      @toc.nil? ? '' : @toc
+
+    @markup.toc = toc_str
+
+    data.gsub(/\[\[_TOC_(.*?)\]\]/) do
```
bartkamphorst: Could we replace this `gsub` call with `gsub!` without issues? If so, then we should do it.

lib/gollum-lib/filter/toc.rb
```diff
@@ -24,15 +25,38 @@ def process(data)
...
+      levels = nil
+
+      opts = Regexp.last_match[1].split('|')[1..-1] || []
+      opts.each do |attr|
```
bartkamphorst: Seems like this is quite an expensive operation if all we want is the value of levels. I think a regexp would be much faster, or not?

bartkamphorst:
No objections to the functionality, just some minor points which might improve performance.

bartkamphorst:
Ping @ztlpn .

ztlpn:
Yeah, sorry. Should have done this earlier.

bartkamphorst:
Cool. @dometto Feel free to merge!

dometto:
Thanks @ztlpn!

PR #188, opened 2015-12-27
Abstract revert functionality

dometto:
Simplify the implementation of `Wiki#revert_page` by abstracting low-level grit calls to the grit adapter (https://github.com/gollum/grit_adapter/pull/14). I've manually tested that this branch with the corresponding branch of `grit_adapter` behaves the same as `master`, but would appreciate if someone else could verify this (since the revert tests don't work with grit).

### DIFF
--- a/lib/gollum-lib/wiki.rb
+++ b/lib/gollum-lib/wiki.rb
@@ -488,7 +488,7 @@ module Gollum
       multi_commit ? committer : committer.commit
     end
 
-    # Public: Applies a reverse diff for a given page.  If only 1 SHA is given,
+    # Public: Applies a reverse diff for a given page. If only 1 SHA is given,
     # the reverse diff will be taken from its parent (^SHA...SHA).  If two SHAs
     # are given, the reverse diff is taken from SHA1...SHA2.
     #
@@ -510,10 +510,8 @@ module Gollum
         sha2   = nil
       end
 
-      patch     = full_reverse_diff_for(page, sha1, sha2)
       committer = Committer.new(self, commit)
-      parent    = committer.parents[0]
-      committer.options[:tree] = @repo.git.apply_patch(parent.sha, patch)
+      committer.options[:tree] = @repo.git.revert(page ? page.path : nil, sha1, sha2, committer.parents[0])
       return false unless committer.options[:tree]
       committer.after_commit do |index, _sha|
         @access.refresh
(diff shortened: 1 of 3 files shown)


dometto:
Note: to test this you have to manually add the following to `Gemfile`:

```ruby
gem `gollum-grit_adapter`, :git => 'https://github.com/gollum/grit_adapter/', :branch => 'abstract_revert'
```

And remove the `grit_adapter` dependency from `gemspec.rb`.

lib/gollum-lib/wiki.rb
```diff
@@ -510,10 +510,8 @@ def revert_page(page, sha1, sha2 = nil, commit = {})
         sha2   = nil
       end
 
-      patch     = full_reverse_diff_for(page, sha1, sha2)
       committer = Committer.new(self, commit)
-      parent    = committer.parents[0]
-      committer.options[:tree] = @repo.git.apply_patch(parent.sha, patch)
+      committer.options[:tree] = @repo.git.revert(page ? page.path : nil, sha1, sha2, committer.parents[0])
```
bartkamphorst: I'm not sure I understand this behavior here, as the method is called `revert_page`. So I would assume that this method has to be called with page object. There is a separate `revert_commit` method below, which should do something else than revert_page.

dometto: Yes, but `revert_commit` just calls `revert_page` with `nil` as the `page` argument. If we're changing the API anyway we can try to make this less counterintuitive. What about these three methods:
* private: `revert(page = nil, sha1, sha2, commit = {})` (implements current logic of `revert_page`)
* public: `revert_page(page, sha1, sha2, commit = {})` (calls `revert` with a page argument)
* public: `revert_commit(sha1, sha2, commit = {})` (calls revert_commit)

dometto:
Refactred according to your suggestions. Would be great if you could do a manual test to verify this behaves the same as the current version of `gollum-lib`.

dometto:
Also, I guess this is a breaking change. Should I reopen this against a new `5.x` branch?

lib/gollum-lib/wiki.rb
```diff
@@ -922,5 +857,62 @@ def tree_map_for(ref, ignore_page_file_dir=false)
     def inspect
       %(#<#{self.class.name}:#{object_id} #{@repo.path}>)
     end
+
+    private
+
+    # Public: Applies a reverse diff for a given page, or an entire commit if page is nil. If only 1 SHA is given,
```
bartkamphorst: Is this still Public?

bartkamphorst:
> Also, I guess this is a breaking change. Should I reopen this against a new 5.x branch?

That would be good, yes.

lib/gollum-lib/wiki.rb
dometto: Ah no, thanks.

bartkamphorst:
Okay, I just tested out the `simplify_revert` branch with the `abstract_revert` grit adapter and reverting pages works without a hitch. :+1:

PR #189, opened 2015-12-30
Only parse tilde based fenced code blocks in markdown. Fixes #171.

### LINKED ISSUE #171: AsciiDoc h3 headers are broken
First off, sorry for the onslaught of AsciiDoc-related Issues...

Asciidoctor includes backwards compatibility with old AsciiDoc headers. Unfortunately, kramdown-style code blocks use conflicting syntax, so this will break rendering.

```
Header
~~~~~~ // should become an h3 header
```

(Even more fun, I just discovered that using kramdown-style code block syntax inside of GFM code blocks breaks GitHub comments. -.-)

I propose simply [turning off](https://github.com/gollum/gollum-lib/blob/master/lib/gollum-lib/filter/code.rb#L25) kramdown-style code blocks when `@markup.format == :asciidoc`. Any objections?

bartkamphorst:


dometto:
:+1:

dometto:
*sigh* I'm inclined to just disable `1.9.3` in travis. ;)

PR #190, opened 2015-12-31
Simplify revert

dometto:
Simplify the implementation of Wiki#revert_page by abstracting low-level grit calls to the grit adapter (gollum/grit_adapter#14). I've manually tested that this branch with the corresponding branch of grit_adapter behaves the same as master, but would appreciate if someone else could verify this (since the revert tests don't work with grit).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,3 +1,2 @@
 source 'https://rubygems.org'
-
 gemspec :name => RUBY_PLATFORM == 'java' ? 'gollum-lib_java' : 'gollum-lib'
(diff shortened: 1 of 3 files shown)


bartkamphorst:
I just tested out the `simplify_revert` branch with the `abstract_revert` grit adapter and reverting pages works without a hitch. :+1:

dometto:
Just pushed grit-adapter 2.0 and made sure this branch uses that dependency. That should fix the build.

PR #193, opened 2016-02-02
Fixes #158

### LINKED ISSUE #158: Test failing on Arch Linux / Ruby 2.2.0
While investigating #152 and #153 , I found [a test from master](https://github.com/gollum/gollum-lib/blob/master/test/test_hook.rb#L93) was failing on my arch box, but not on my ubuntu box. At the request of @dometto, I am opening a separate issue.

```bash
[nathan@nathan-arch gollum-lib]$ bundle exec rake test
/usr/bin/ruby -I"lib:lib:test:."  "/usr/lib/ruby/2.2.0/rake/rake_test_loader.rb" "test/**/test_*.rb" 
Loaded suite /usr/lib/ruby/2.2.0/rake/rake_test_loader
Started
.....................................................................................................................................................
...................F
=====================================================================================================================================================
Failure: test_push_and_pull(Pushing_and_pulling)
/home/nathan/projects/gollum-lib/test/test_hook.rb:93:in `block (2 levels) in <top (required)>'
<false> expected but was
<true>

diff:
? false
? tru  
=====================================================================================================================================================
.................................................................................................

Finished in 3.552655921 seconds.
-----------------------------------------------------------------------------------------------------------------------------------------------------
266 tests, 603 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
99.6241% passed
-----------------------------------------------------------------------------------------------------------------------------------------------------
74.87 tests/s, 169.73 assertions/s
rake aborted!
Command failed with status (1): [ruby -I"lib:lib:test:."  "/usr/lib/ruby/2.2.0/rake/rake_test_loader.rb" "test/**/test_*.rb" ]

Tasks: TOP => test
(See full trace by running task with --trace)
```

natejenkins:
Adds sleep in test_push_and_pull to ensure ordering of commits.

On Ruby 2.2.4, this test was failing for me.  Looking at the clone git repo from the command line the ordering of commits was correct and included the second commit, however inspecting with byebug had the commits in the wrong order.  I'm pretty sure this is due to them both having the same timestamp.  Looking into the Commit class in gitlab-grit shows that it uses rev_list when calculating the array of commits, and according to https://git-scm.com/docs/git-rev-list: 

```
git-rev-list - Lists commit objects in reverse chronological order
```

The simplest fix is to include a one second sleep between the commits, which I have done here and now all tests are passing.  It would be possible to properly order the commits when the timestamp is equal, but that would require modifying gitlab-grit and my goal here was simply to get all tests passing in gollum-lib.

### DIFF
--- a/test/test_hook.rb
+++ b/test/test_hook.rb
@@ -90,7 +90,7 @@ context 'Pushing and pulling' do
     @origin.write_page("Gollum2", :markdown, "New content2", {:message => "Wrote second test page"})
     @clone.repo.git.pull("origin", "master")
     # Rugged does not support high-level pull yet, so pull is implemented as a merge. Hence need to check for merge commit on rugged adapter.
-    assert_equal ["Wrote second test page", "Merged branch refs/heads/master of origin."].include?(@clone.repo.commits.first.message), true
+    assert_equal ["Wrote second test page", "Merged branch refs/heads/master of origin."].include?(@clone.repo.head.commit.message), true
   end
   
   teardown do


bartkamphorst:
Thanks @natejenkins for addressing this issue (and for your excellent work at authorea!). I appreciate the simplicity of the proposed fix, but I am hesitant to include `sleep` statements in the tests as they typically only mask an underlying problem and unnecessarily slow down the test suite. Do you see any other way of fixing the issue?

bartkamphorst:
I restarted the build, but there are still failures. We've had issues with 1.9.3 on Travis before, so I'm not too worried about that. Any idea about the jruby-build failing though?

Thanks for the alternative fix, I see no problems with looking at the head of the repo.

bartkamphorst:
> Any idea about the jruby-build failing though?

Never mind, this really is unrelated to the PR at hand.

bartkamphorst:
As a final request, would you mind squashing the two commits?

bartkamphorst:
Nope, go ahead!

bartkamphorst:
Thanks! :+1:

PR #194, opened 2016-02-10
GlobalTOC: Fix relpath issue when navigating in a subdirectory

vaussard:
The GlobalTOC macro uses relative paths to navigate. This causes an
issue if the user is located in a subdirectory, as this path will be
added to the current location and point to an invalide page.

A simple fix is to use absolute paths by prefixing with '/'.

### DIFF
--- a/lib/gollum-lib/macro/global_toc.rb
+++ b/lib/gollum-lib/macro/global_toc.rb
@@ -3,7 +3,7 @@ module Gollum
     class GlobalTOC < Gollum::Macro
       def render(title = "Global Table of Contents")
         if @wiki.pages.size > 0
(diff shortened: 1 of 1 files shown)


vaussard:
Ok I also need to update the corresponding test in gollum-lib/test/test_macros.rb:44. Before doing this, is there other issues?

dometto:
On 10-02-16 12:59, Florian Vaussard wrote:
>
> Ok I also need to update the corresponding test in 
> gollum-lib/test/test_macros.rb:44. Before doing this, is there other 
> issues?
>

Thanks for the PR. I think this is fine, but yes, a test would be very 
useful.

vaussard:
My patch is making the current test to fail, so I will repost with an updated test.

PR #195, opened 2016-02-10
GlobalTOC: Fix relpath issue when navigating in a subdirectory (2)

vaussard:
The GlobalTOC macro uses relative paths to navigate. This causes an
issue if the user is located in a subdirectory, as this path will be
added to the current location and point to an invalide page.

A simple fix is to use absolute paths by prefixing with '/'.


vaussard:
Now the CI fails for 2 jobs on 5, but it seems not due to my patch.

dometto:
Indeed, test failures seem to be due to something else, so merging this. Thanks!

PR #196, opened 2016-02-19
Fix TOC with level behavior after #176

dometto:
Merging #176 apparently caused the new behavior introduced by #183 to fail. #176 cleaned up the TOC html, but the new max-levels functionality was expecting the less clean html. In particular, [this check](https://github.com/gollum/gollum-lib/blob/master/lib/gollum-lib/filter/toc.rb#L50) was no longer accurate: #176 added `<li>` tags surrounding the `<ul>` so that `ancestors.length` became larger.

I've attempted to solve the issue by looking at *only* the ancestors that are `<ul>`'s to ascertain the level of a heading. I also had to update the test html.

@ztlpn: would appreciate a quick glance at my fix to `toc.rb` if you could spare the time.


ztlpn:
@dometto Thanks! The fix looks good.

PR #197, opened 2016-02-23
Put back MRI 2.2 and 2.3 in Travis CI

jhominal:
The underlying issue that had led to the deactivation
of these versions of Ruby was Issue #158, which was solved
by PR #193.


bartkamphorst:
:+1: Thanks.

PR #198, opened 2016-02-24
Improves efficiency of subpage lookup.

bartkamphorst:
@dometto @jhominal I'm open to suggestions for further improvements!

### DIFF
--- a/lib/gollum-lib.rb
+++ b/lib/gollum-lib.rb
@@ -3,6 +3,7 @@
 require 'digest/md5'
 require 'digest/sha1'
 require 'ostruct'
+require 'pathname'
 
 DEFAULT_ADAPTER = RUBY_PLATFORM == 'java' ? 'rjgit_adapter' : 'grit_adapter'
 
--- a/lib/gollum-lib/file.rb
+++ b/lib/gollum-lib/file.rb
@@ -1,5 +1,4 @@
 # ~*~ encoding: utf-8 ~*~
-require 'pathname'
 
 module Gollum
   class File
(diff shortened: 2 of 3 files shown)


dometto:
I get that it's better to look for all the subpage types in one go, but ideally, we would not need to loop through (or even generate) the entire tree map -- just the trees underlying the current path. That would probably require some changes to `Wiki#tree_map_for` (or a new helper method), but I wonder whether that wouldn't be worth the effort, if we're rewriting this method anyway?

(Or maybe I'm not getting the logic and the entire treemap is not being looped through?)

dometto:
Otherwise clean code, btw! Have you tried any benchmarks?

bartkamphorst:
> Have you tried any benchmarks?

It roughly shaves off one second or so. Not a great improvement, but still.

bartkamphorst:
> ideally, we would not need to loop through (or even generate) the entire tree map -- just the trees underlying the current path.

You got the logic right, and yes, I can see how this might improve things further. Suggestions for improving `Wiki#tree_map_for`?

dometto:
I was thinking of an optional `path` argument, which on the lower level should then trigger an `ls_tree` for that path rather than for the entire repo.

dometto:
I just realized that this of course won't improve things for pages in the root of the repository. If subpage lookup for pages in the root turns out to be a major performance impediment, we should maybe think about requiring the user to either put the subpages in the same directory as the page, or in a `subpage` directory in the same directory as the page.

bartkamphorst:
@dometto Shall we just merge this?

dometto:
Sorry, missed your ping there. Yeah, just merge it. :)

bartkamphorst:
👍

PR #199, opened 2016-02-28
Support utf-8 page names unless using Grit

hiroponz:
### Problem

It needs that apps using gollum-lib override String class to support utf-8 page names.


hiroponz:
Sorry, this change breaks gollum.

PR #200, opened 2016-03-04
Configure plantuml filter via environment variable.

hsanson:
Modify plantuml filter to support configuration via environment variables. This is useful when running gollum from the command line.

    export PLANTUML_URL="https://my-plantuml-server:9091/plantuml/png"
    gollum

### DIFF
--- a/lib/gollum-lib/filter/plantuml.rb
+++ b/lib/gollum-lib/filter/plantuml.rb
@@ -37,7 +37,7 @@ require 'zlib'
 #
 class Gollum::Filter::PlantUML < Gollum::Filter
 
-  DEFAULT_URL = "http://localhost:8080/plantuml/png"
+  DEFAULT_URL = ENV["PLANTUML_URL"] || "http://localhost:8080/plantuml/png"
 
   # Configuration class used to change the behaviour of the PlatnUML filter.
   #


bartkamphorst:
I like the idea of making this configurable, but I would prefer to do it via the `config` file instead of ENV. Would you agree @dometto ?

dometto:
It should already be doable via config.rb, right? Though I guess that would throw a warning about redefining the constant.

I'm not against this if there's a reason why ENV is more convenient, but otherwise I'd prefer not too have unnecessarily many places where configuration is defined. @hsanson, can you elaborate on the need for this as opposed to using config.rb?

hsanson:
When running gollum from a terminal to work on a local wiki (e.g. gilab wiki), this is the easiest way I found to configure gollum-lib without modifying the source. If gollum can be configured by other means (e.g. config.rb?) please let me know.

    > cd my-wiki
    > PLANTUML_URL=http://my-plantuml-server/png  gollum

hsanson:
By the way this can be configured using the following snippet too:

```
require "gollum-lib"

Gollum::Filter::PlantUML.configure do |config|
    config.url = "http://my-own-plantuml.com:8080/plantuml/png"
end
```

See the wiki: https://github.com/gollum/gollum/wiki/Custom-PlantUML-Server#configuring-gollum-lib

dometto:
Try the `--config path/to/config.rb` option. You can just put your snippet in there (but you don't need the `require 'gollum-lib'`.

hsanson:
@dometto I just tested using the --config option and it works as expected. I will update the wiki with this information.

I still prefer using environment variables for quick, one time, wiki edits on my local machine and would be
glad if this pull request gets accepted. Is just a one-line change and does not affect the confing.rb functionality.

bartkamphorst:
@hsanson Thanks for updating the wiki! :+1: I am going to close this PR, however, since the `config.rb` file is gollum's preferred route for personalization.

PR #202, opened 2016-04-14
Rugged adapter as default

KHN190:
Turned to rugged adapter to solve utf-8 title. Rugged seemed stable now.


dometto:
This adapter is not ready to become the default. See e.g. https://github.com/gollum/rugged_adapter/issues/15

PR #203, opened 2016-04-15
Support for emoji

svoop:
See https://github.com/gollum/gollum/issues/1125 for more.

### DIFF
--- a/lib/gollum-lib/filter/emoji.rb
+++ b/lib/gollum-lib/filter/emoji.rb
@@ -4,20 +4,36 @@
 #
 # Render emoji such as :smile:
 class Gollum::Filter::Emoji < Gollum::Filter
-  EMOJI_TAG_PATTERN = %r{
+
+  EXTRACT_PATTERN = %r{
     (?<!\[{2})
-    :(?:([\w-]|</?em>)+):
+    :(?<name>[\w-]+):
     (?!\]{^2})
   }ix
 
+  PROCESS_PATTERN = %r{
+    =EEMMOOJJII=
+    (?<name>[\w-]+)
+    =IIJJOOMMEE=
+  }ix
+
   def extract(data)
+    data.gsub! EXTRACT_PATTERN do
+      emoji_exists?($~[:name]) ? "=EEMMOOJJII=#{$~[:name]}=IIJJOOMMEE=" : $&
+    end
     data
   end
 
   def process(data)
-    data.gsub EMOJI_TAG_PATTERN do |tag|
-      name = tag[1..-2].gsub(%r{</?em>}, '_')   # undo emphasis
-      %Q(<img src="/emoji/#{name}" alt="#{name}" class="emoji">)
-    end
+    data.gsub! PROCESS_PATTERN, %q(<img src="/emoji/\k<name>" alt="\k<name>" class="emoji">)
+    data
   end
+
+  private
+
+  def emoji_exists?(name)
+    @index ||= Gemojione::Index.new
+    !!@index.find_by_name(name)
+  end
+
 end
--- a/lib/gollum-lib/filter/emoji.rb
+++ b/lib/gollum-lib/filter/emoji.rb
@@ -1,0 +1,23 @@
+# ~*~ encoding: utf-8 ~*~
+
+# Emoji
+#
+# Render emoji such as :smile:
+class Gollum::Filter::Emoji < Gollum::Filter
+  EMOJI_TAG_PATTERN = %r{
+    (?<!\[{2})
+    :(?:([\w-]|</?em>)+):
+    (?!\]{^2})
+  }ix
+
+  def extract(data)
+    data
+  end
+
+  def process(data)
+    data.gsub EMOJI_TAG_PATTERN do |tag|
+      name = tag[1..-2].gsub(%r{</?em>}, '_')   # undo emphasis
+      %Q(<img src="/emoji/#{name}" alt="#{name}" class="emoji">)
+    end
+  end
+end
--- a/gemspec.rb
+++ b/gemspec.rb
@@ -8,7 +8,7 @@ def specification(version, default_adapter, platform = nil)
     s.name              = 'gollum-lib'
     s.version           = version
     s.platform          = platform if platform
-    s.date              = '2016-02-21'
+    s.date              = '2016-04-15'
     s.rubyforge_project = 'gollum-lib'
     s.license           = 'MIT'
 
@@ -30,6 +30,7 @@ def specification(version, default_adapter, platform = nil)
     s.add_dependency 'stringex', '~> 2.5.1'
     s.add_dependency 'sanitize', '~> 2.1.0'
     s.add_dependency 'github-markup', '~> 1.4.0'
+    s.add_dependency 'gemojione', '~> 2.2'
 
     s.add_development_dependency 'org-ruby', '~> 0.9.9'
     s.add_development_dependency 'kramdown', '~> 1.6.0'
@@ -70,6 +71,7 @@ def specification(version, default_adapter, platform = nil)
     lib/gollum-lib/file_view.rb
     lib/gollum-lib/filter.rb
     lib/gollum-lib/filter/code.rb
+    lib/gollum-lib/filter/emoji.rb
     lib/gollum-lib/filter/macro.rb
     lib/gollum-lib/filter/metadata.rb
     lib/gollum-lib/filter/plain_text.rb
--- a/lib/gollum-lib.rb
+++ b/lib/gollum-lib.rb
@@ -15,6 +15,7 @@ end
 # external
 require 'github/markup'
 require 'sanitize'
+require 'gemojione'
 
 # internal
 require File.expand_path('../gollum-lib/git_access', __FILE__)
@@ -61,4 +62,3 @@ module Gollum
   class NoSuchPathError < StandardError; end
 
 end
-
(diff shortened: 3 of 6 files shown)


lib/gollum-lib/filter/emoji.rb
```diff
@@ -0,0 +1,23 @@
...
+    (?!\]{^2})
+  }ix
+
+  def extract(data)
+    data
+  end
+
+  def process(data)
+    data.gsub EMOJI_TAG_PATTERN do |tag|
+      name = tag[1..-2].gsub(%r{</?em>}, '_')   # undo emphasis
```
bartkamphorst: Could you explain why this line is here? Where does the emphasis come from?

lib/gollum-lib/filter/emoji.rb
```diff
@@ -0,0 +1,23 @@
...
+    :(?:([\w-]|</?em>)+):
+    (?!\]{^2})
+  }ix
+
+  def extract(data)
+    data
+  end
+
+  def process(data)
+    data.gsub EMOJI_TAG_PATTERN do |tag|
```
bartkamphorst: Could you use `gsub!` instead, or does the return value need to be non-nil? Same question for the second call to `gsub`.

lib/gollum-lib/filter/emoji.rb
svoop: Take the emoji `:do_not_litter:`. Since the markdown parser kicks in first, by the time the emoji is parsed, it will be `:do<em>not</em>litter:`. Of course, this collision could be solved properly elsewhere, but since emoji are an opt-in feature, I figured this should be taken care of in the context of the emoji even if it appears a little hackish.

lib/gollum-lib/filter/emoji.rb
svoop: It should work with `gsub!` as well, however, what's the advantage of it in this case (and even more so in the one on the following line)?

lib/gollum-lib/filter/emoji.rb
bartkamphorst: Actually this is why the filter chain is traversed twice (the second time in reverse). You could extract `name` in the `extract` method and replace that string with a temporary placeholder string that is not in danger of being parsed by another filter (the tag filter uses `"TAG#{Digest::SHA1.hexdigest(tag)}TAG"`). Then, in `process` you could replace that placeholder with the HTML string (and the `name` variable should still be available). Take a look at how the tags filter works for an example.

dometto: > On 23 mei 2016, at 15:11, Sven Schwyn <notifications@github.com> wrote:
> 
> Of course, this collision could be solved properly elsewhere, but since emoji are an opt-in feature, I figured this should be taken care of in the context of the emoji even if it appears a little hackish.

I agree with that approach. Just mind that, since the filter chain is customizable, we cant assume that the markdown parser has already kicked in. No problem for this code, just good to keep in mind that users *can* possibly break this functionality if they're not careful. :)

dometto: On 23 mei 2016, at 17:21, Bart Kamphorst <notifications@github.com<mailto:notifications@github.com>> wrote:

Actually this is why the filter chain is traversed twice (the second time in reverse). You could extract name in the extract method and replace that string with a temporary placeholder string that is not in danger of being parsed by another filter (the tag filter uses "TAG#{Digest::SHA1.hexdigest(tag)}TAG"). Then, in process you could replace that placeholder with the HTML string (and the name variable should still be available). Take a look at how the tags filter works for an example.

Err, yes, that approach fixes the problem I just pointed out. :)

lib/gollum-lib/filter/emoji.rb
bartkamphorst: It's just that `gsub` makes a copy of the string it operates on, so it slightly increases memory footprint. And anything we can do to spare the garbage collector will benefit speed.

lib/gollum-lib/filter/emoji.rb
bartkamphorst: Actually that is why the filter chain is traversed twice (the second time in reverse). You could extract `name` in the `extract` method and replace it with a temporary placeholder string that is not in danger of being parsed by another filter (the tags filter uses `"TAG#{Digest::SHA1.hexdigest(tag)}TAG"`). Then, on the second pass, you can replace that placeholder string with your HTML snippet in the `process` method (`name` should still be available). Take a look at the tags filter for an example.

svoop: @bartkamphorst I'll take a look at this, however, can't do it today, my schedule is a little tight.

lib/gollum-lib/filter/emoji.rb
svoop: Okidoke, I'll change that. (Might backfire though once Ruby goes immutable strings.)

lib/gollum-lib/filter/emoji.rb
bartkamphorst: 👍

dometto:
Waiting for the last small changes to merge this :+1:

svoop:
Oh, sorry, slipped from my radar... I'll do it this evening.

svoop:
@dometto There you go with to GC-friendly `gsub!`s. Wonder if it really makes a difference though.

dometto:
Thanks! I think there's just the matter of extracting the em-tags on the first pass of the filter chain then. :)

svoop:
Can you give me a hint there?

bartkamphorst:
This is the point about the filter chain being traversed twice. The idea would be to use `gsub!` in `extract(data)` to replace emoji tags (using `EMOJI_TAG_PATTERN`) with a unique placeholder that is not at risk of being parsed by another filter. Then in `process(data)` you would replace that placeholder again with the HTML pointing to the emoji icon. That way you can get rid of the "undo emphasis" bit. Let me know if that explanation makes sense to you.

svoop:
@bartkamphorst Yes, absolutely, I'll implement it like so.

bartkamphorst:
Any luck @svoop ? There's no rush, but we've been asked by the folks at Gitlab to produce a new release, and it would be nice to include emoji support. But we could also schedule another release later on.

svoop:
@bartkamphorst I'll be hitting the road tomorrow and I'm therefore a little bound by the things to prepare. But I can get this done early next week. Can you wait until then before you cut a release?

svoop:
Almost there, I'll wrap it up this evening, got to run now.

bartkamphorst:
Excellent. Thanks for the update!

svoop:
@bartkamphorst I've switched to the two-step approach you've suggested. I'm sorry for the delay!

bartkamphorst:
Very clean, thank you! Happy with this as well, @dometto ? As a final request @svoop , would you mind squashing your 4 commits into one?

dometto:
Happy with this too! And yes, please squash. Thanks @svoop!

svoop:
I'll sqash them this evening... on the road right now.
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

svoop:
@dometto I've squashed the last three commits, however, I can't include the fourth since there's a merge commit in between and things would break. So there's are two commits in the history now.

svoop:
@dometto @bartkamphorst 

I've just stumbled over an edge-case: Since the "extract" doesn't check whether an emoji actually exists, the following Rake codeblock example doesn't render nicely since `:schema:` is treated as if it were an emoji:

    bundle exec rake db:mongoid:create

Result:

> screen shot 2016-06-30 at 01 44 50

I can easily fix this by checking the EmojiOne index and only replace existing emoji names while leaving everything else as is. This seems like a good idea to me, but the lookup comes at a small price in terms of CPU and MEM load.

What do you prefer, more speed or more accuracy? I'd opt for the latter.

dometto:
Yes, more accuracy please! :) also be sure the emojis are placed in the filter chain such that code blocks will be extracted earlier.

svoop:
@dometto WILCO – I'll have some spare time this evening.

svoop:
@dometto Only emoji present on the GemojiOne index are extracted now,

> also be sure the emojis are placed in the filter chain such that code blocks will be extracted earlier

I played around with the position in the filter chain, but no matter where I place `:Emoji`, emoji inside code blocks are always interpreted.

https://github.com/gollum/gollum-lib/pull/203/commits/b214c41b50384919382d0657b72399b4c502e915#diff-078dc3a04ea60d229edd0ea0e1213435L253

connorshea:
gemojione 3.0.0 is going to come out in the next few days, figured I should mention that.

dometto:
@svoop, in your opinion, would it be worth to include gemojione `v3.0.0`?

svoop:
Sure, why not. I'll do it tomorrow morning.
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

svoop:
@dometto I've upgraded to Gemojione 3, only a minor change was necessary both on gollum and gollum-lib. However, since you have already merged the "emoji" branch over at gollum, would you prefer to merge here first and then I do fresh PRs on both gems for the Gemojione 3 update?

dometto:
Yes, we'll do it that way. Merging now, and looking forward to the `gemojione` update!

connorshea:
I opened an issue for upgrading to/allowing 3.x: #226

PR #213, opened 2016-08-07
Allow use of HTTPS for plantuml. 

dometto:
See https://github.com/gollum/gollum/issues/1117. @j4hangir  @hsanson, could you give this PR a spin and let me know if it resolves the issue satisfactorily? (you can `git checkout -b fix_plantuml_https origin/fix_plantuml_https`)

### DIFF
--- a/lib/gollum-lib/filter/plantuml.rb
+++ b/lib/gollum-lib/filter/plantuml.rb
@@ -45,10 +45,11 @@ class Gollum::Filter::PlantUML < Gollum::Filter
   #   test: Set to true when running tests to skip the server check.
   #
   class Configuration
-    attr_accessor :url, :test
+    attr_accessor :url, :test, :verify_ssl
 
     def initialize
       @url = DEFAULT_URL
+      @verify_ssl = true
       @test = false
     end
   end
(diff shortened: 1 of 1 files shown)


hsanson:
I don't have a plantuml server with HTTPS but I can say this PR does not break my setup. If the PR has been tested to work with HTTPS then I see no problem in merging this.

dometto:
Ping @j4hangir, mind giving this a go?

j4hangir:
@dometto Sure, but please be patient :)

j4hangir:
I replaced `/var/lib/gems/1.9.1/gems/gollum-lib-4.1.0/lib/gollum-lib/filter/plantuml.rb` with your file and changed the protocol to HTTPS, but the very same error is shown.

dometto:
The problem was likely that SSL certificate checking failed. This is not something we can easily fix inside gollum, but you may be able to configure ruby's `net/http` library in such a way that it recognizes the certificate (by setting a custom CA file). Otherwise, I've added an option to skip certificate checking for PlantUML settings. See the wiki:
https://github.com/gollum/gollum/wiki/Custom-PlantUML-Server#ssl

I've now succesfully tested the use of PlantUML under https (using the server at https://plantuml.org/plantuml/png), and will merge this once the tests succeed.

PR #215, opened 2016-08-14
Fix subpage lookup bug. Fixes #214.

### LINKED ISSUE #214: Broken _sidebar.md
With the changes from PR #198 , files named `_sidebar.md` no longer load (They need to be named `_Sidebar.md` - case-sensitive to load correctly with the changes).  That is a minor issue - but breaks previous functionality.

Also - subdirectories do not load the correct `_Sidebar.md`.  For example, given the following structure:

 * root
    * index.md
    * _Sidebar.md
    * subdir
        * index.md
        * _Sidebar.md

If you load `index.md`, the top-level `_Sidebar.md` is loaded.  However, if you load `subdir/index.md`, the top-level `_Sidebar.md` is still loaded.

I narrowed it down to commit 2fb3bea by rolling back one commit at a time between version 2.4.0 and 2.4.1 of `gollum-lib`.

I, personally, haven't noticed a huge improvement in page load time with #198 added - and it just breaks these cases.  Perhaps it could be rolled back?

bartkamphorst:
This fixes the case described by @toonetown in #214 . It does not reintroduce case insensitiveness, as that was never intended behavior (all the docs use `_Sidebar.md` etc.).  I'm curious to see how this solution will perform on a large repository.

### DIFF
--- a/lib/gollum-lib/page.rb
+++ b/lib/gollum-lib/page.rb
@@ -477,17 +477,31 @@ module Gollum
       
       map ||= @wiki.tree_map_for(@wiki.ref, true)
       valid_names = subpagenames.map(&:capitalize).join("|")
-      while entry = map.shift do
-        next unless entry.name =~ /^_(#{valid_names})/
-        sub_page_type = ::File.basename(entry.name.downcase.tr('_', ''), ::File.extname(entry.name))
-        next if instance_variable_get("@#{sub_page_type}")
-        ::Pathname.new(::File.join('.', ::File.dirname(self.path))).ascend do |dir|
-          if ::File.join(dir, entry.name) == ::File.join('.', ::Pathname.new(entry.path))
-            instance_variable_set("@#{sub_page_type}", entry.page(@wiki, @version) )
+      # From Ruby 2.2 onwards map.select! could be used
+      map = map.select{|entry| entry.name =~ /^_(#{valid_names})/ }
+      return if map.empty?
+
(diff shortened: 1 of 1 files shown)


dometto:
No real comments, looks clean! Also wondering about the performance. Have you tried profiling it?

The previous implementation had a `next if instance_variable_get(subpagename)` to speed up one of the loops -- not sure if there is ever actually a case in which the subpage lookup is called while one of the subpages is already set , but if there is, would it help to have something similar in this implementation?

Oh, and do we need to improve the test suite to deal with @toonetown's cases?

bartkamphorst:
> Have you tried profiling it?

Only by looking at total page load times of a repo with 4000 commits. There doesn't seem to be a big speed difference there (if anything, the new implementation seems to be slightly faster, unless the path is highly nested).

> The previous implementation had a next if instance_variable_get(subpagename) to speed up one of the loops -- would it help to have something similar in this implementation?

Different logic. That case does not occur in this implementation.

> Oh, and do we need to improve the test suite to deal with @toonetown's cases?

Ideally yes, but I don't think I'll have time to update the test suite on short notice. If you want to have a stab at it, please do!

Rejown:
👍

bartkamphorst:
I'm in favor of merging this into master, since it fixes two bugs that were introduced with the previous implementation of `find_sub_pages`. Okay @dometto ?

dometto:
Sure, go ahead!

PR #220, opened 2016-09-04
Speed up page loading via caching name canonicalization

roccomoretti:
Page loading on our Gollum wiki is very slow. In part that's due to the large number of page and having a fair number of links per page. 

Investigation with various profiling utilities  (e.g. ruby-prof) indicated that the majority of the time was spent in text processing, particularly in page and file name canonicalization. The same text normalization steps were being done for every page/file in the repository, for each link being rendered. 

To fix this, I've taken the caching scheme from the GitAccess object and added it to Wiki object. In the Wiki object the appropriately pre-processed page and file names are now cached, on a per-SHA1 basis. This pre-processed data gets stored in a Hash, so the rendering of each link after the very first should be sped up significantly.

On my local machine, for my particular wiki of interest, launching Gollum, rendering 4 example pages and exiting out takes a total of ~22 CPU seconds ("user" time) with current master. Using the changes in this pull request, this is reduced to ~1.9 CPU seconds (a 10-fold speedup).

---

One issue I'm still a bit hazy on is the memory impact on long-running processes. I think I have set things up such that the cached values in the Wiki object get cleared out occasionally, but I'm not confident enough in how long-running Gollum systems work to say that the new scheme won't slowly accumulate memory as additional SHA1s get added.

### DIFF
--- a/lib/gollum-lib/file.rb
+++ b/lib/gollum-lib/file.rb
@@ -140,7 +140,9 @@ module Gollum
       map     = @wiki.tree_map_for(version)
       commit  = version.is_a?(Gollum::Git::Commit) ? version : @wiki.commit_for(version)
 
-      if (result = map.detect { |entry| entry.path.downcase == checked })
+      file_map = @wiki.tree_file_map_for(version)
+      if file_map.has_key?(checked)
+        result = file_map[checked]
         @path    = name
         @version = commit
 
--- a/lib/gollum-lib/page.rb
+++ b/lib/gollum-lib/page.rb
@@ -353,9 +353,13 @@ module Gollum
     #
     # Returns the String canonical name.
     def self.cname(name, char_white_sub = '-', char_other_sub = '-')
-      name.respond_to?(:gsub) ?
-          name.gsub(%r(\s), char_white_sub).gsub(%r([<>+]), char_other_sub) :
-          ''
+      if name.respond_to?(:tr) and char_white_sub.size == 1 and char_other_sub.size == 1
+          name.tr( " \t\r\n\f" '<>+', ('\\'+char_white_sub)*5 + ('\\'+char_other_sub)*3 )
+      else
+          name.respond_to?(:gsub) ?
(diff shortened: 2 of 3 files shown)


dometto:
Thanks for the PR! This sounds great. I'll have a closer look ASAP.

Meanwhile, I've also been working on improving performance in gollum itself, see my last three comments in https://github.com/gollum/gollum/issues/1078. Cumulatively, these changes should make significant dents in the load times. (Although only this PR addresses some of the fundamental issues in gollum-lib)

bartkamphorst:
Very promising speedups! Ideally we also look into improving the underlying code, but caching does seem desirable here. Let's test this for inclusion in the 5.x release.

roccomoretti:
Just wanted to double check that there isn't anything I still needed to do in order to move this pull request forward.

bartkamphorst:
Actually, it would be great if you could open this PR against the `gollum-lib-5.x` branch instead of `master`. Do you agree, @dometto ?

If you happen to have used this improvement in your own environment, do you have any more insights into how it might affect long-running Gollum instances?

dometto:
Yes, agreed! Sorry for the much too slow response, and thanks for the great work. Let's discuss further when you reopen the PR.

bartkamphorst:
@roccomoretti Could you please base this PR against the `gollum-lib-5.x` branch?

dometto:
Ping @roccomoretti

dometto:
Actually, in `5.x` we've completely removed page name canonicalization as it's no longer necessary. That sort of obviates the need for this kind of caching. So the choice is whether to merge this into `master` or not at all.

bartkamphorst:
Not at all in that case; only bugfixes to master at this point as far as I'm concerned.

dometto:
@roccomoretti thanks for the excellent PR, but as you can see I'm afraid we should have come up with something like this a couple of years ago. :) At least when `5.x` comes out you can have the benefit of much faster page loads.

roccomoretti:
@dometto That's perfectly fine. I unfortunately haven't been able to find the free time to give this the attention it needed. -- But as long as page loading got faster, I'm happy.

PR #221, opened 2016-09-06
Retain the `@wiki.tree_map_for()` cache

Rejown:
When I use the `<<Navigation()>>` macro in a `_Sidebar`, the `_Sidebar` preview can show the page list, contains all the pages in the same directory, but other pages which include the sidebar will only show an empty page list.

the reason is, `map.shift` will clear the tree map cache, if the `find_sub_pages` executed, the cache of the tree map turned into an empty array, and will not be updated, because of an empty array is not nil.

### DIFF
--- a/lib/gollum-lib/page.rb
+++ b/lib/gollum-lib/page.rb
@@ -477,7 +477,7 @@ module Gollum
       
       map ||= @wiki.tree_map_for(@wiki.ref, true)
       valid_names = subpagenames.map(&:capitalize).join("|")
-      while entry = map.shift do
+      map.each do |entry|
         next unless entry.name =~ /^_(#{valid_names})/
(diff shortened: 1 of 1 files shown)


bartkamphorst:
@Rejown Thanks for the PR. There is a proposal for an alternative implementation of find_sub_pages [here](https://github.com/gollum/gollum-lib/pull/215), which I think would resolve your issue as well. Could you take a look and let me know whether you agree?

Rejown:
Thanks for attention. Your implementation looks great, I expect #215 to be merged. 
I will close this, thank you.

bartkamphorst:
Thanks for checking out the implementation. I just merged the new `find_sub_pages` method into master, so your issue should be fixed in the next release. Apologies for breaking functionality in the first place!

PR #229, opened 2016-12-12
fix bug:can not support files with non-ascii name and passed

greatbody:
this pull request passed local test

fix bug:can not support files with non-ascii name


greatbody:
can't pass jruby.don't knwo why...

PR #230, opened 2016-12-15
update and try pass jruby test

greatbody:
update and try pass jruby test

### DIFF
--- a/gemspec.rb
+++ b/gemspec.rb
@@ -8,7 +8,7 @@ def specification(version, default_adapter, platform = nil)
     s.name              = 'gollum-lib'
     s.version           = version
     s.platform          = platform if platform
-    s.date              = '2016-11-18'
+    s.date              = '2016-12-11'
     s.rubyforge_project = 'gollum-lib'
     s.license           = 'MIT'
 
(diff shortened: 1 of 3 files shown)


dometto:
Thanks for the PR. However, the jruby errors on travis do not seem related to encoding issues, but rather to bundler/gemspec failures -- or is there an additional issue you were trying to address with this? If so, please let me know, and I'll reopen this!

greatbody:
Hi, long time since last commit.
I created this PR to solve a problem that affects Chinese uses.
At the time I created this PR, gollum cannot let use create files using Chinese,and would transfer Chinese To English charactors.

In my PR, I make it possible to create markdown with Chinese title.
I believe Japenese can do that too if my PR was merged.

Thanks!

bartkamphorst:
@greatbody Thanks for the PR, but we want to tackle UTF-8 encoding in a more robust way in gollum version `5.X`. Keep an eye out for that release; it should allow Chinese page names without transliteration.

greatbody:
@bartkamphorst Thanks for your reply, I will be watching.

Issue #226, opened 2016-11-20
Update gemojione to 3.x


connorshea:
Right now the gemspec locks [gemojione](https://github.com/jonathanwiesel/gemojione) to `~> 2.2`, it'd be nice to have 3.x support in gollum-lib.

The main problem is that there's no support for Ruby 1.9 in gemojione 3.0. Either 1.9 would need to be dropped or we could allow either 2.x/3.x and use separate gemfiles to run tests with gemojione 2.x and 3.x.

bartkamphorst:
I think dropping Ruby 1.9 for the 5.x release of gollum certainly is an option. Do you agree @dometto ?

svoop:
@connorshea I'll take care of this.

svoop:
@connorshea Oh, didn't see the Ruby 1.9 constraint. IMO, it's no good to add unneccessary complexity to the code for the sake of supporting a Ruby which reached it's EOL almost two years ago. I'd rather couple the Emoji-Update to dropping Ruby 1.9 as @bartkamphorst suggested.

dometto:
> I'd rather couple the Emoji-Update to dropping Ruby 1.9 as @bartkamphorst suggested.

👍

In that case, I propose to also get rid of [these lines](https://github.com/gollum/gollum/blob/master/lib/gollum.rb#L14-L16) :)

svoop:
@dometto In order to prepare a PR for Gollum 5 with the update to gemojione 3: Are the branches `gollum-lib:gollum-lib-5.x` and `gollum:5.x` the baselines to use?

dometto:
Correct, @svoop!

svoop:
There you go, sorry for the delay. As expected, the jruby-19mode tests fail.

connorshea:
@svoop can we close this now?

svoop:
@connorshea i'd say yes

PR #231, opened 2016-12-28
Update gemojione to 3.2

svoop:
https://github.com/gollum/gollum-lib/issues/226


svoop:
The jruby-19mode test fails because Gemojione ~> 3 has dropped support for Ruby 1.9. See https://github.com/gollum/gollum-lib/issues/226 for the discussion.

PR #232, opened 2016-12-30
Issue #1161 : Fix rendering of TOC on the sidebar.

bogado:
All the subpages were not setting the "parent_page" field correctly. So
the TOC was operating directly in the subpage instead of the actual
page.

### DIFF
--- a/lib/gollum-lib/page.rb
+++ b/lib/gollum-lib/page.rb
@@ -6,6 +6,7 @@ module Gollum
     Wiki.page_class = self
     
     SUBPAGENAMES = [:header, :footer, :sidebar]
+    @@to_xml       = { :save_with => Nokogiri::XML::Node::SaveOptions::DEFAULT_XHTML ^ 1, :indent => 0, :encoding => 'UTF-8' }
     
     # Sets a Boolean determing whether this page is a historical version.
     #
@@ -119,22 +120,8 @@ module Gollum
     # This is usually equal to either the first h1.
     #
     # Returns the title extracted from the content.
-    #
-    # TODO: this case is a bad encapsulation, the markup
-    # class should be able to "know" how to extract this.
     def content_title
-        case format
-          when :asciidoc
-            doc.css("h1:first-child")
-          when :org
-            doc.css("p.title:first-child")
-          when :pod
-            doc.css("a.dummyTopAnchor:first-child + h1")
-          when :rest
-            doc.css("div > div > h1:first-child")
-          else
-            doc.css("h1:first-child")
-        end
+        find_header_node(@doc).inner_text.strip
     end
 
     # Public: The title will be constructed from the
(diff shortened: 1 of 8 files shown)


lib/gollum-lib/page.rb
```diff
@@ -466,6 +506,28 @@ def page_match(name, path)
       false
     end
 
+    # Private: get the node where the header is located
+    #
+    # Extracts the header node from doc.
+    #
+    # TODO: this case is a bad encapsulation, the markup
```
dometto: Agreed, we don't want markup-specific code here. How much work would it be to change that?

bogado: I say that the formaters should own this knowledge, so this should simply forward the call to the correct formater. Another option would be to have a special class to handle the title concern, I would prefer this, as the concerns seem to be loosely distribute along different classes. A normal thing to happen when projects have many devs and evolve more or less chaotic.

On my opinion the formater have a loose contract with the rest of the wiki in terms of html generation. I understand the "why" this is like that, so that the assimilation of new formats is simpler. But I think this is the root cause of the problem here. If the rest of the system could have more certain that this would simply be the the first `h1` or that it would be marked as class/id `title` for instance, this code would be simpler.

dometto:
@bogado looks like this PR is addressing multiple issues at once, and getting it ready to merge will be very difficult because of recent changes to the `5.x` branch. If I'm not mistaken the actual fix for TOC on subpages is only very small, though. Could you open a new PR, or modify this one, to address just that issue?

Other issues, like refactoring the page title generation, are also very important and we would very much welcome your input on that, as well! But I would prefer to do that in separate PRs.

karlredman:
- Unfortunately, it looks like the [suggested fix](https://github.com/gollum/gollum-lib/pull/232/commits/e19b4f5363d84f64beb53bbc29468d75b693ee1d) from @bogado never made it into a release.
- I'm adding a note here for people who want to just get the functionality of TOC in their _Sidebar/_Footer files while a "proper implementation" is sorted out by devs.
- I have confirmed that this fix works as a monkeypatch for gollum v4.1.2 (which uses gollum-lib v5.0.a.3)
    - for an example of a monkey patched golllum config.rb see [here](https://github.com/karlredman/Vimwiki-Gollum-Integration/blob/master/mockwiki/gollum_admin/config.rb)

dometto:
Ping @bogado  -- still interested in getting the bugfix merged?

dometto:
> I have confirmed that this fix works as a monkeypatch for gollum v4.1.2 (which uses gollum-lib v5.0.a.3) 

For the record, I strongly advise against using the alpha release of `gollum-lib`, and certainly not in combination with a `4.x` release of `gollum`. `gollum-lib 5.x` is inteded for use future use with `gollum 5.x`.

karlredman:
Thank you for taking another look. It would be awesome if the fix from @bogado makes in in somewhere. Also, thank you for the help/correction with my pairing gollum v4.x with gollum-lib v5.x. I'm not sure why I didn't catch that major version numbers were completely different. doh! I have made notes this effect in my own projects. Thanks again!

dometto:
@karlredman you're also more than welcome to open a new PR that just contains the fix (and any adapted tests)!

karlredman:
Hi, I've submitted a pull re-request. I do feel like I should probably add more information but I'm not sure. Let me know if I should add anything else. Thank you again for your patience.

PR #233, opened 2016-12-31
Dev

bogado:
This makes the page object responsible for deriving the correct title based on the h1_title wiki option.

The logic here was removed from the gollum app, where it was being processed in the view.

### DIFF
--- a/gemspec.rb
+++ b/gemspec.rb
@@ -26,9 +26,9 @@ def specification(version, default_adapter, platform = nil)
 
     s.add_dependency *default_adapter
     s.add_dependency 'rouge', '~> 2.0'
-    s.add_dependency 'nokogiri', '~> 1.6.4'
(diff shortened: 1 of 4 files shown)


dometto:
@bogado thanks for the PR. I think we'd still very much like to have this improvement. However, could you:
* reopen it against the `gollum-lib-5.x` branch?
* take out the version updates for the dependencies, unless necessary for this fix (otherwise we'd rather have them as a separate PR)
* possibly take out the `Macro` fixes, which should also be a separate PR (unless it's necessary to make the tests pass)
* address the test failures

dometto:
Ping @bogado

bartkamphorst:
@bogado Are you still interested in working on this? Otherwise I'm closing the PR for now.

bartkamphorst:
Closing due to lack of activity.

PR #235, opened 2017-01-14
Make navigation macro strip root path correctly

ProgramFan:
The `navigation` macro does not strip root path correctly when the path contains multiple words, while it works correctly for single word root paths. For example:

`<<navigation("Research Topic 1", "Research-Topic-1")>>` gives something like

    Research Topic 1/Ideas 1
    Research Topic 1/Ideas 2

while it should be

    Ideas 1
    Ideas 2

This PR fixes it.

### DIFF
--- a/lib/gollum-lib/macro/navigation.rb
+++ b/lib/gollum-lib/macro/navigation.rb
@@ -6,7 +6,7 @@ module Gollum
         if @wiki.pages.size > 0
           list_items = @wiki.pages.map do |page|
             if page.url_path.start_with?(toc_root_path)
-              path_display = full_path ? page.url_path_display  : page.url_path_display.sub(toc_root_path,"").sub(/^\//,'')
+              path_display = full_path ? page.url_path_display  : page.url_path_display.sub(toc_root_path.gsub("-", " "), "").sub(/^\//,'')
               "<li><a href=\"/#{page.url_path}\">#{path_display}</a></li>"
             end
           end


lib/gollum-lib/macro/navigation.rb
```diff
@@ -6,7 +6,7 @@ def render(title = "Navigate in the TOC", toc_root_path = ::File.dirname(@page.p
         if @wiki.pages.size > 0
           list_items = @wiki.pages.map do |page|
             if page.url_path.start_with?(toc_root_path)
-              path_display = full_path ? page.url_path_display  : page.url_path_display.sub(toc_root_path,"").sub(/^\//,'')
+              path_display = full_path ? page.url_path_display  : page.url_path_display.sub(toc_root_path.gsub("-", " "), "").sub(/^\//,'')
```
dometto: Could you explain how this addresses the issue? If `url_path_display` is something like `Research Topic 1/Ideas 1`, why should we `gsub` all dashes (`-`)?

dometto:
Tests failed on jruby, but should be fine now once we restart it.

lib/gollum-lib/macro/navigation.rb
ProgramFan: The problem is the `toc_root_path`, which is dashed as `Research_Topic_1` in this case. So I undash it to make the stripping work as expected. Note that for single-word `toc_root_path` such as "ResearchTopic1", there is no problem.

dometto: Ah I see, it's another whitespace-substitution issue. We're working on removing all whitespace->dash substitution in `v5.x`, because it's unnecessary and leads to a lot of bugs and oversights like this. @bartkamphorst, would you agree we merge this into `master` as a bugfix though?

bartkamphorst: 👍

dometto:
Sorry, forgot to merge this before releasing `4.2.4`. It will be in the next bugfix release then. @bartkamphorst: when we merge `5.x` into `master` we should be careful about this one, as it reintroduces as ws sub.

dometto:
One build on Travis seems stuck for some reason. @bartkamphorst could you merge this later today if the test ends up passing?

bartkamphorst:
All lights are green. Merging.

PR #238, opened 2017-01-20
Whitespace in page URL paths is now changed to dashes

dblessing:
Previously, Gollum::Page#url_path would change whitespace in the
page *name* portion of the path to dashes. However, whitespace
in the path itself was retained. This caused problems in URLs.

This was originally created as a bug in GitLab at https://gitlab.com/gitlab-org/gitlab-ce/issues/26831, for reference. 

I hope I went about adding an example to the lotr.git repo correctly. Admittedly, I'm not a LOTR fan 😱  (sorry) so I had to ask a coworker to help me come up with a new item. He suggested 'Dol Guldur' and 'Sauron' - I hope this fits with the theme. 

Thanks for your review.


dometto:
Thanks for the PR! The test failure is because of a general problem with testing jruby on travis we're experiencing, so I will merge regardless. The only question I have is whether there is a scenario in which this might break existing wikis. If so, I'd like to merge into a different branch than `master`. Or was it simply impossible to have whitespace in the path until now? 

And yes, your coworker did well. :)

dblessing:
> The only question I have is whether there is a scenario in which this might break existing wikis.

Great question. I tested before and after in `gollum` proper and both work. `gollum` itself seems to honor both `Dol%20Guldur` and `Dol-Guldur` in the URL, though I'm not sure how that works. I tested the patch in GitLab and it works after the patch. 

I don't *think* it will break anything, but I can't be 100% sure.

dblessing:
@dometto What do you think?

dometto:
Since the tests pass, I'm willing to merge it into `master`, unless @bartkamphorst objects.

bartkamphorst:
Actually, I prefer to merge this into the `5.x` branch, just to be on the safe side. @dblessing Would you mind opening this PR against that branch?

dometto:
Hmm, but now in the `5.x` branch we're working on removing all whitespace->dash conversion. Shouldn't we just consider this a bugfix for versions `< 5` and *not* merge it into `5.x`?

bartkamphorst:
I'm not in favor of doing so because we don't know whether it will break existing behavior.

dometto:
In that case perhaps we should just close this? Sorry for the trouble @dblessing: the story is we want to get rid of the whitespace conversion because it leads to endless bugs like the one you caught, and serves no real purpose anymore (it is an artifact of using `grit`).

PR #241, opened 2017-02-09
Allow gollum installs with different root URLs to delete files

zachfeldman:
Update file_view.rb to allow gollum installations with different root urls than / to delete files by making the deleteFile path relative.

### DIFF
--- a/lib/gollum-lib/file_view.rb
+++ b/lib/gollum-lib/file_view.rb
@@ -25,7 +25,7 @@ module Gollum
     end
 
     def delete_file(url, valid_page)
-      %Q(<form method="POST" action="/deleteFile/#{url}" onsubmit="return confirm('Do you really want to delete the file #{url}?');"><button type="submit" name="delete" value="true"></button></form>)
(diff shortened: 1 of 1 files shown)


bartkamphorst:
Could you elaborate on your use case? Also, this produces six errors in the test suite which will need to be fixed before we can merge.

zachfeldman:
So, I'm mounting Gollum at a /wiki route instead of the normal root path route. When I did that, I noticed that the file delete url was an absolute URL from / instead of relative, like most of this project's URLs.

I'll close the PR since this is pretty specific to my use case. But maybe it will help somebody else in the future!

bartkamphorst:
I hadn't realized that the delete route deviated from the others. I'm in favor of handling all routes uniformly, so we might just re-implement your fix. 👍

zachfeldman:
@bartkamphorst cool! Glad to advise but don't have tons of time for the patch. Good luck :D

dometto:
>  I'm in favor of handling all routes uniformly, so we might just re-implement your fix.

So this PR should be merged in combination with a change of the router `gollum`? The failing tests should also be fixed.

PR #248, opened 2017-03-31
Update gemspec.rb

rogerhu:
Upgrade nokogiri


bartkamphorst:
Thanks for the PR. I will not merge this into master, however, since we're waiting with updating nokogiri until version 5.x, because the new nokogiri version no longer supports older rubies, and that constitutes a breaking change for gollum.

PR #249, opened 2017-04-01
Remove ws subs

dometto:
Got rid of almost all test errors and failures on this branch (9 failures, 1 error remaining). To fix the remaining test failures and errors (see Travis) we must change the expected output and thus make some decisions:

* Should we remove special chars and or HTML from filenames? See `test_page.rb:103`
* Getting pages case sensitive or not? See `test_page.rb:33`
* match against file extension or not? See `test_page.rb:94`
* what should Page#canonicalize_filename do when presented with a non-string or nil argument? At present I return '' for any item that is not of class 'String'.


dometto:
Also, we should think what`Page#url_path` should return in case of filepaths containing whitespace: URL encode the string (so whitespace -> `+`)?

Gemfile
```diff
@@ -1,3 +1,4 @@
 source 'https://rubygems.org'
 
 gemspec :name => RUBY_PLATFORM == 'java' ? 'gollum-lib_java' : 'gollum-lib'
+gem "gollum-rugged_adapter"
```
bartkamphorst: Should this gem dependency be in this PR?

bartkamphorst:
> Should we remove special chars and or HTML from filenames?

I may be missing a use case, but feel that in general this kind of sanitation should really happen as early in the chain as possible, so in gollum, and not gollum-lib. But even if we keep in in gollum-lib, I think we should return `nil` instead of trying to coerce it to some page title that otherwise matches.

bartkamphorst:
> Getting pages case sensitive or not?

Not, so _case insensitive_. This is the simplest option, and will work across filesystems (and ruby 2.4 actually supports downcasing utf-8 chars now 👍). I don't think there is pertinent use case for having two pages with the exact same title, but with mixed case instead of lowercase.

bartkamphorst:
> match against file extension or not?

Yes. I think all valid pages should have a file extension on the file system. If it is left out of the request (so 'Home' instead of 'Home.md'), gollum-lib will try to return the first page that matches (could be 'Home.asciidoc' if that also happens to exist), but it should in principle only serve pages (and perhaps also files) that have a file extension.

bartkamphorst:
> what should Page#canonicalize_filename do when presented with a non-string or nil argument? At present I return '' for any item that is not of class 'String'.

I think the pattern`filename.is_a?(String)` should generally be avoided. Better to coerce to a String by calling `to_s` if `filename.respond_to?(:to_s)` (and otherwise throw an error, because that really shouldn't be happening).

dometto:
> Yes. I think all valid pages should have a file extension on the file system. If it is left out of the request (so 'Home' instead of 'Home.md'), gollum-lib will try to return the first page that matches (could be 'Home.asciidoc' if that also happens to exist), but it should in principle only serve pages (and perhaps also files) that have a file extension.

I also think that would be cleanest. That means at least:
* not stripping the extension from `canonicalized_filename`
  * but make sure the page title doesn't show the extension
* a fallback routine in `Page.find_page_in_tree` (strip extension from `haystack` if `needle` is also extensionless)
* check all the write methods for where they assume the page title is extension-less

dometto:
> (and otherwise throw an error, because that really shouldn't be happening).

Fair enough. Curious to see what happens if we raise an exception there... :)

bartkamphorst:
> Also, we should think whatPage#url_path should return in case of filepaths containing whitespace: URL encode the string (so whitespace -> +)?

Again, I wonder if Page objects really need to have this responsibility of turning paths into a URLs, or whether this is something that gollum frontend (or another frontend) should take care of.

bartkamphorst:
> Fair enough. Curious to see what happens if we raise an exception there... :)

`NilClass` might turn out to be a special case, depending on how the method is used, but otherwise there shouldn't be any objects passed to that method that cannot be coerced to a String.

dometto:
> I wonder if Page objects really need to have this responsibility of turning paths into a URLs, or whether this is something that gollum frontend (or another frontend) should take care of.

Agreed, but I think this PR should leave the current way of doing things mostly intact in order to make the tests pass. Given that [the filter chain requires a Page#escaped_url_path method](https://github.com/gollum/gollum-lib/blob/master/lib/gollum-lib/filter/tags.rb#L291) we should provide that method for now. We can perhaps come up with a cleaner way of doing things in a following PR.

bartkamphorst:
> not stripping the extension from canonicalized_filename but make sure the page title doesn't show the extension

Unless `canonicalized_filename` is only called when trying to get a page ready for rendering, in that case stripping the extension is fine as we already know by then what kind of file it is. Or is the method also called before lookup?

bartkamphorst:
> Agreed, but I think this PR should leave the current way of doing things mostly intact in order to make the tests pass.

Yes, I'm totally on board with that!

So in that case it is probably a good start to replace spaces with '+' signs in `url_path`.

Gemfile
dometto: I wanted to switch to `rugged` in order to make sure that everything would work without the ws subs. I'll check if the tests pass with the `grit` adapter in the end, though. If not, shall I move this dependency to the gemspec instead?

dometto:
> Or is the method also called before lookup?

Unfortunately, yes. For instance [here](https://github.com/gollum/gollum-lib/pull/249/files#diff-9c66db1000504e8fd86544f23961803eR276) and [here](https://github.com/gollum/gollum-lib/pull/249/files#diff-e5bff0f40f8bf41e5d5572ec29834e53R393).

However, `canonicalize_filename` now basically only a) calls `stripped_filename` and b) checks whether the argument is a string. Given our present discussion, we would stop doing a), and b) would just amount to calling `to_s`. So perhaps we could remove this method altogether? Or do we like the idea of having all filename construction methods use a canonicalization method, even if it doesn't do much at present?

lib/gollum-lib/committer.rb
```diff
@@ -87,8 +87,8 @@ def parents
     # Returns nothing (modifies the Index in place).
     def add_to_index(dir, name, format, data, allow_same_ext = false)
       # spaces must be dashes
-      dir.gsub!(' ', '-')
-      name.gsub!(' ', '-')
+      #dir.gsub!(' ', '-')
+      #name.gsub!(' ', '-')
```
dometto: Will remove the commented out lines.

lib/gollum-lib/page.rb
```diff
@@ -147,15 +146,16 @@ def url_path
...
+      path << self.name
       path
     end
 
     # Public: The display form of the url path required to reach this page within the repo.
     #
     # Returns the String url_path
     def url_path_display
-      url_path.gsub("-", " ")
+      # REMOVED GSUB
```
dometto: Will remove the placeholder comments.

lib/gollum-lib/wiki.rb
```diff
@@ -436,7 +421,7 @@ def update_page(page, name, format, data, commit = {})
       dir      = ::File.dirname(page.path)
       dir      = '' if dir == '.'
       filename = (rename = page.name != name) ?
-          Gollum::Page.cname(name) : page.filename_stripped
+          name : page.filename_stripped
```
dometto: This is inconsistent at present: I switched to using `canonicalize_filename` later during constructing this PR. However, as I mentioned, the result is at present the same. So we could remove the `canonicalize_filename` method and just use `filename_stripped` where necessary.

bartkamphorst: Right. However, as you mentioned, perhaps this PR should be restricted to only removing whitespace replacement. 😄

lib/gollum-lib/page.rb
bartkamphorst: 👍

lib/gollum-lib/committer.rb
bartkamphorst: 👍

bartkamphorst:
> Or do we like the idea of having all filename construction methods use a canonicalization method, even if it doesn't do much at present?

If we have absolutely no reason for it, there is also no reason for the extra method call. However, we need to think about cases in which canonicalization is desirable, given that we will no longer convert utf-8 to ascii and spaces to dashes. Maybe @sunny could weigh in on this?

dometto:
Just pushed some new changes:
* case insensitivity restored
* removed the html or special chars in filename tests
* allows matching against file extension
* improved performance of `Page#valid_page_name?` and the like

lib/gollum-lib/page.rb
```diff
@@ -389,10 +382,10 @@ def find_page_in_tree(map, name, checked_dir = nil, exact = false)
       name        = ::File.join(checked_dir, name) if checked_dir
 
       map.each do |entry|
-        next if entry.name.to_s.empty?
-        entry_name = Page.canonicalize_filename(entry.name)
+        next if entry.name.to_s.empty? || !self.class.valid_extension?(entry.name)
+        entry_name =  ::File.extname(name).empty? ? Page.canonicalize_filename(entry.name) : entry.name
         path = checked_dir ? ::File.join(entry.dir, entry_name) : entry_name
-        next unless name == path
+        next unless name.downcase == path.downcase
```
dometto: Can't compare `name.downcase! == path.downcase!` as the `!` methods return `nil` when the string are already downcase. But perhaps:
```ruby
path.downcase!
name.downcase!
next unless name == path
```
?

Or should we not get into this habit anyway as ruby moves toward making string immutable by default?

bartkamphorst: I think your snippet with the bang methods is right for now. When ruby 3.0 rolls around we'll have to go through the whole code base anyway.

dometto:
I've managed to fix the test failures except for two, which fail for the same reason. Apparently gollum expects the following markdown (from `Bilbo-Baggins.md`)...
```
Bilbo Baggins is the protagonist of The [[Hobbit]] and also makes a few
appearances in The Lord of the Rings, two of the most well-known of [[J. R. R.
Tolkien]]
```
to render as:
```
Bilbo Baggins is the protagonist of The <a class=\"internal present\" href=\"/Hobbit\">Hobbit</a> and also makes a few\nappearances in The Lord of the Rings, two of the most well-known of <a class=\"internal absent\" href=\"/J.+R.+R.+Tolkien\">J. R. R.\nTolkien</a>
```
Note the newline in the middle of the link tag, which the tests thus expect gollum to ignore -- because such newlines were removed by calling the `Page.cname` method.

With the current implementation, the link renders as ` <a class=\"internal absent\" href=\"/J.+R.+R.%0ATolkien\">J. R. R.\nTolkien</a>`: the newline is URL-encoded.

I see two ways to go:
* prevent the tag filter from matching tags including newlines (so changing gollum's behavior)
* sanitizing the newlines
  * specifically where link tags are renderer, in the tag filter's `process_link_tag` method
  * everywhere where we now removed `Page.cname` calls, so basically reintroduce that method 

I *would* prefer the first method: I don't see a good use case for tags containing newlines inside them. However, it does look like github's wikis support multi-line tags. :-(

dometto:
Ah yes, interestingly, removing the ws subs works even with the `grit_adapter` and the current `gitlab-grit` implementation. So I've removed the switch to rugged from this PR.

dometto:
> For instance:
[Wikipe
dia]
renders as a link with the textWiki pedia (newline sub'd for whitespace).

Given this I would prefer to also sub newlines for whitespace inside the link text.

Apart from this the question remains at which point to sanitize input such that we don't end up writing filenames with newline characters to git.

dometto:
Correction: `[[gollum|style]]` tags on github wiki's containing newline do not seem to render as links. So I propose we also stop matching newlines within tags.

dometto:
Tests are now passing. I don't know why Travis isn't testing on jruby, but I've just manually tested that it works on that platform too. When we merge this we should probably merge recent commits from `master` ASAP to maintain mergability between the two branches.

lib/gollum-lib/page.rb
```diff
@@ -17,29 +17,25 @@ class Page
...
+    # ext - String file extension, like ".md"
     #
-    # Returns e.g. ["Home", :markdown], or [] if the extension is unregistered
-    def self.parse_filename(filename)
-      return [] unless filename =~ /^(.+)\.([a-zA-Z]\w*)$/i
-      pref, ext = Regexp.last_match[1], Regexp.last_match[2]
-
+    # Returns the Symbol format associated with ext
+    def self.parse_extension(ext)
       Gollum::Markup.formats.each_pair do |name, format|
```
dometto: This is still not very efficient: we have to loop through all the registered markups and check whether the defined regexp matches `ext`. I don't actually understand why we don't simply register valid extensions as strings rather than regexps, so that we could just do a `VALID_EXTENSIONS.include?(ext)` here. Perhaps this is good further refactoring for `5.x`?

bartkamphorst: We want to know the `name` corresponding to the matching extension, not just whether there is a match. What we could do is rewrite `register` to generate a hash like this:
```ruby
formats = {
  'md': {name: :markdown},
  'mdown': {name: :markdown},
  'asciidoc': {name: :asciidoc},
  'creole': {name: :creole, reverse_links: true},
}
```
That way, `formats[ext]` would give us all we want to know.

bartkamphorst:
Terrific work, @dometto . 👍 Have you tried running gollum in the browser with this branch of `gollum-lib`?

dometto:
Yes, using this branch of `gollum-lib`, `gollum` runs successfully. I am able to get pages of the form `Super duper` and `Super duper.md`, however, the front-end seems to still perform its own whitespace substitution when attempting to create new pages. So a PR will be necessary for `gollum`, too.

The `gollum` tests pass successfully, except for two failures which are caused by the tests' excepting ws substitution.

Shall I merge this into `5.x` then?

bartkamphorst:
Yes, let's go for it!

PR #251, opened 2017-04-06
Detect wiki formats availability

ProgramFan:
Gollum registers all known wiki format, regardless of whether or not the corresponding gems are installed. This confuses users sometimes, as seen in [gollum/gollum#705](https://github.com/gollum/gollum/issues/705). This PR fixes it by prefiltering formats.

### DIFF
--- a/lib/gollum-lib/markups.rb
+++ b/lib/gollum-lib/markups.rb
@@ -31,31 +31,28 @@ module Gollum
       Kramdown::Document.new(content, :auto_ids => false, :input => "markdown").to_html
     }
 
-    # markdown, rdoc, and plain text is always supported.
-    register(:markdown,    "Markdown", :regexp => /md|mkdn?|mdown|markdown/)
-    register(:rdoc,        "RDoc")
-    register(:txt,         "Plain Text")
-
-    if MarkupRegisterUtils::gem_exists? "RedCloth"
-      register(:textile,   "Textile")
-    end
-    if MarkupRegisterUtils::gem_exists? "org-ruby"
-      register(:org,       "Org-mode")
-    end
-    if MarkupRegisterUtils::gem_exists? "creole"
-      register(:creole,    "Creole", :reverse_links => true)
-    end
-    if MarkupRegisterUtils::executable_exists? "python2"
-      register(:rest,      "reStructuredText", :regexp => /re?st(\.txt)?/)
-    end
-    if MarkupRegisterUtils::gem_exists? "asciidoctor"
-      register(:asciidoc,  "AsciiDoc")
-    end
-    if MarkupRegisterUtils::gem_exists? "wikicloth"
-      register(:mediawiki, "MediaWiki", :regexp => /(media)?wiki/, :reverse_links => true)
-    end
-    if MarkupRegisterUtils::executable_exists? "perl"
-      register(:pod,       "Pod")
-    end
+    # markdown, rdoc, and plain text are always supported.
+    register(:markdown, "Markdown", :regexp => /md|mkdn?|mdown|markdown/)
+    register(:rdoc, "RDoc")
+    register(:txt, "Plain Text")
+    # the following formats are available only when certain gem is installed
+    # or certain program exists.
+    register(:textile, "Textile",
+             :enabled => MarkupRegisterUtils::gem_exists?("RedCloth"))
+    register(:org, "Org-mode",
+             :enabled => MarkupRegisterUtils::gem_exists?("org-ruby"))
+    register(:creole, "Creole",
+             :enabled => MarkupRegisterUtils::gem_exists?("creole"),
+             :reverse_links => true)
+    register(:rest, "reStructuredText",
+             :enabled => MarkupRegisterUtils::executable_exists?("python2"),
+             :regexp => /re?st(\.txt)?/)
+    register(:asciidoc, "AsciiDoc",
+             :enabled => MarkupRegisterUtils::gem_exists?("asciidoctor"))
+    register(:mediawiki, "MediaWiki",
+             :enabled => MarkupRegisterUtils::gem_exists?("wikicloth"),
+             :regexp => /(media)?wiki/, :reverse_links => true)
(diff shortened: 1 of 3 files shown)


lib/gollum-lib/markups.rb
```diff
@@ -1,20 +1,61 @@
 # ~*~ encoding: utf-8 ~*~
 module Gollum
+  module MarkupRegisterUtils
+    # Check if a gem exists. This implementation requires Gem::Specificaton to
+    # be filled.
+    def gem_exists? name
+      return Gem::Specification.find {|x| x.name == name} != nil
+    end
+
+    # Check if an executable exsits. This implementation comes from
```
bartkamphorst: typo: exists.

ProgramFan: Fixed.

ProgramFan:
Unfortunately, this commit break the tests "test_Gollum_Markup_formats_returns_all_formats_by_default". So any ideas on this?

dometto:
Thanks for the PR!

I would be ok with changing the expected output of that test (and renaming it), since we're precisely trying to stop gollum from showing all markups by default.

What consequences does this approach have for links to pages that are in a format for which no markup gem is currently installed? Currently those are recognised as valid, present pages. Are they now treated as "absent" or, as Files that are present? I assume the latter. That may be OK new behavior, but it would be a breaking change, so this would have to go into the gollum-lib-5.x branch.

bartkamphorst:
What about just making uninstalled markup options red in the dropdown menu? That way there is an intuitive way for users to see that certain markups _could_ be available, but are not (and this is different from when markups are explicitly limited via `config.rb`).

ProgramFan:
I understand your concerns. A better approach is that we add the format availability check here in gollum-lib and make use of that information to mark uninstalled markup options in the editor in special style. I will do it.

ProgramFan:
Done in the latest commit. The change shall be backward compatible now. The latest commit to gollum/gollum#1173 make use of this information to visually signal unavailable formats, also in a backward compatible way.

lib/gollum-lib/markups.rb
```diff
@@ -1,20 +1,58 @@
 # ~*~ encoding: utf-8 ~*~
 module Gollum
+  module MarkupRegisterUtils
+    # Check if a gem exists. This implementation requires Gem::Specificaton to
+    # be filled.
+    def gem_exists? name
```
bartkamphorst: Style-wise, I prefer parentheses around method arguments (`gem_exists?(name)`) just to keep it consistent with the rest of the gollum code base.

lib/gollum-lib/markups.rb
```diff
@@ -1,20 +1,58 @@
...
+      return Gem::Specification.find {|x| x.name == name} != nil
```
bartkamphorst: Explicit return statement is not needed.
Also, please choose a more descriptive variable name than `x` like `gem` or `gemspec`.

lib/gollum-lib/markups.rb
```diff
@@ -1,20 +1,58 @@
...
+    end
+
+    # Check if an executable exists. This implementation comes from
+    # stackoverflow question 2108727.
+    def executable_exists? name
+      exts = ENV["PATHEXT"] ? ENV["PATHEXT"].split(";") : [""]
```
bartkamphorst: What does PATHEXT contain exactly?

lib/gollum-lib/markups.rb
```diff
@@ -1,20 +1,58 @@
...
+      paths = ENV["PATH"].split(::File::PATH_SEPARATOR)
+      paths.each do |path|
+        exts.each do |ext|
+          exe = ::File.join(path, "#{name}#{ext}")
```
bartkamphorst: Please use `Pathname` API here.

lib/gollum-lib/markups.rb
ProgramFan: Done.

bartkamphorst:
@ProgramFan This is really terrific, thanks for working on this! 👍

lib/gollum-lib/markups.rb
ProgramFan: Done

lib/gollum-lib/markups.rb
ProgramFan: It shall be something like ".exe" on windows, and nothing on Linux/macOS. But I have not tried that on windows, just copied that from the SO answer.

lib/gollum-lib/markups.rb
ProgramFan: Good idea. Done.

ProgramFan:
Jruby failed for no reason, but all other CI checks succeeded.

dometto:
I just restarted the jruby build and it fails again. I think there may be a real issue here with the attempt to check whether certain files exists or are executable.

dometto:
Perhaps you could try updating `.travis.yml` so that we use a more recent `jruby`: change `jruby-19mode` to `jruby-9.1.8.0`.

ProgramFan:
All tests passed after the jruby change. Why?

dometto:
The test failure was probably caused by a bug in older implementations of `jruby`. Pathname/file operations is one area where `jruby` used to be unreliable, at least before the release of `jruby-9k`.

At least it works now! The question is whether we should consider this a breaking change if we can't get it to work on `jruby-19mode`. Arguably not, since `gollum` already requires `ruby >= 2.0` on MRI. @bartkamphorst?

bartkamphorst:
I was going to propose to merge this into `5.x` instead of master anyway (I prefer to only merge necessary bug fixes into master at this point), so in that case it is okay if this PR introduces breaking changes.

dometto:
@ProgramFan looks like the most pragmatic thing to do would be to reopen this against the`gollum-lib-5.x` branch, then we can merge it right away. Seems like a better option than trying to fix the failures on old versions of `jruby`...

ProgramFan:
See #256.

PR #252, opened 2017-04-07
Yaml frontmatter

dometto:
Replaces the oldschool `Metadata` filter with YAML Frontmatter, as per https://github.com/gollum/gollum/issues/774.

All tests pass locally. Notably, the `metadata_title` functionality works like a charm.

In theory, this is all that needs doing, but we may choose to add additional functionality to the frontend, in order to [display frontmatter in a table like github does](https://github.com/blog/1647-viewing-yaml-metadata-in-your-documents). I'm personally not sure if this should be done by default (for someone who just wants to use the `metadata_title` feature, getting the table in each document is unwanted). But perhaps we could render it **iff** the metadata contains e.g. `display_metadata: true`.


dometto:
The old metadata filter did some basic sanitation (removing `<` and `>` characters) which is not currently implemented in the YAML filter:
```ruby
YAML.load "---\ntitle: title\n<script>document.write 'hello world';</script> waa: daa\n---"
 => {"title"=>"title", "<script>document.write 'hello world';</script> waa"=>"daa"} 
```
Basically this only becomes an issue when rendering the metadata as part of the page. Do we want to perform the sanitation wherever it is we render the metadata (i.e. in the frontend), or rather here?

lib/gollum-lib/filter/yaml.rb
```diff
@@ -0,0 +1,26 @@
...
+  
+  # Regexp thanks to jekyll
+  YAML_FRONT_MATTER_REGEXP = %r!\A(---\s*\n.*?\n?)^((---|\.\.\.)\s*$\n?)!m
+
+  def extract(data)
+    data.gsub!(YAML_FRONT_MATTER_REGEXP) do
+      @markup.metadata ||= {}
+      begin
+        frontmatter = ::YAML.load(Regexp.last_match[1])
+        frontmatter = frontmatter.respond_to?(:to_h) ? frontmatter.to_h : Hash.new(frontmatter)
```
bartkamphorst: What is the use case for frontmatter that needs to be coerced into a hash via `Hash.new`?

lib/gollum-lib/filter/yaml.rb
```diff
@@ -0,0 +1,26 @@
...
+        @markup.metadata.merge!(frontmatter)
+      rescue Psych::SyntaxError
```
bartkamphorst: Not a big fan of silently ignoring errors. Could we add a `misformed frontmatter` message to the frontmatter hash, so that would show up on the page?

bartkamphorst:
> Do we want to perform the sanitation wherever it is we render the metadata (i.e. in the frontend), or rather here? Or in Page#metadata?

I think it actually makes sense to do it here in the filter.

bartkamphorst:
Great work, @dometto ! With regard to metadata table rendering, I'm very much in favor of implementing that feature. I propose the following:
* an application wide setting that turns metadata table rendering on or off;
* set metadata table rendering to **enabled** by default;
* only render metadata table if it contains more than just `metadata_title`;
* allow rendering of metadata table to be turned off on a page by page basis with `display_metadata: false`.

Thoughts?

dometto:
> an application wide setting that turns metadata table rendering on or off;
set metadata table rendering to enabled by default;
only render metadata table if it contains more than just metadata_title;
allow rendering of metadata table to be turned off on a page by page basis with display_metadata: false

Really like that idea! I'll start working on an implementation.

I just realized we might also leave in the old `metadata.rb` file without actually applying it in the filter chain by default if folks still want to use it. It can actually be used in conjunction with the YAML filter. But perhaps we should just send the message that people should migrate their wikis?

bartkamphorst:
> I just realized we might also leave in the old metadata.rb file 

If we "deprecate" metadata in 5.x, we'll have to wait until 6.x. to remove it. Unless, as you say, we use it in conjunction, but I don't know what the benefits of that would be. Downside would be maintainance.

lib/gollum-lib/filter/yaml.rb
dometto: the `Hash.new` case is to deal with cases like this:
```ruby
2.4.1 :004 > YAML.load("---\ntrue\n---")
 => true  
```
This surprisingly doesn't throw a syntax error: as long as there is no `:` in the YAML block, it is simply parsed as a YAML object and returns a string, boolean, etc.

Calling `Hash.new(obj)` for any object that doesn't respond to `to_h` just results in an empty hash, so this is a way to make sure the metadata isn't accidentally set to an object of a different kind, without resorting to `is_a?(Hash)`. But I also don't think it's very transparent, so very open to other suggestions.

I also just realized there is this edge case, because `Array`s actually have a working `to_h` method :
```ruby
.4.1 :012 > YAML.load("---\n[[foo, bar],[waa,baa]]\n---").to_h
 => {"foo"=>"bar", "waa"=>"baa"} 
```

lib/gollum-lib/filter/yaml.rb
dometto: Yes, good idea. An `errors` keyword with an `Array` as value? That could actually be a place to store errors collected in the rest of the filter chain, too, should we want to.

dometto:
You're right, let's just ditch it.

lib/gollum-lib/filter/yaml.rb
bartkamphorst: I think what threw me was `Hash.new(frontmatter)` as opposed to just have it be `Hash.new`. The Array edge case is interesting, since `to_h` is defined on each array, but will break on a normal array of type `['a', 'b']` with a `TypeError`. So we would need to rescue those if we want to continue using `to_h`. Alternatively, we could check if the YAML loaded object responds to `keys` and `values` methods?

lib/gollum-lib/filter/yaml.rb
bartkamphorst: 👍

dometto:
Switched to `YAML.safe_load` (see [here](https://github.com/ruby/psych/issues/119#issuecomment-12875715)) and using `Sanitize` for HTML sanitation. Refactored as discussed and added error logging.

dometto:
I've added an API for determining whether a `Page`'s metadata should be rendered according to your suggestion above.

It looks like Travis is failing because `YAML.safe_load` is not available on jruby in `1.9` mode. I think safe mode was added in `ruby v.2`. Should we switch to testing `jruby-9k`?

dometto:
I just over optimistically switched to jruby-9k as it passed locally (https://github.com/gollum/gollum-lib/commit/a7d3ed2c0b528f63a1036019a589cd7ad2059329). Reverted because Travis fails miserable:
https://travis-ci.org/gollum/gollum-lib/jobs/220003063

test/test_markup.rb
```diff
@@ -760,52 +760,80 @@ def assert_markup_highlights_code(markup_class, rendered)
...
 
     index = @wiki.repo.index
     index.add("Bilbo-Baggins.md", content)
-    index.commit("Add metadata")
+    index.commit("Add metadata")    
 
     page     = @wiki.page("Bilbo-Baggins")
     rendered = Gollum::Markup.new(page).render
-    assert_html_equal output, rendered
+    assert_equal output, rendered.gsub(/<(\/)?p>/,'')
```
dometto: Had to insert the ugly `gsub`s because `kramdown` on `jruby` inserts `<p>` around the result (and not on MRI).

dometto:
Green lights 👍

lib/gollum-lib/filter/yaml.rb
```diff
@@ -10,10 +10,11 @@ def extract(data)
     data.gsub!(YAML_FRONT_MATTER_REGEXP) do
       @markup.metadata ||= {}
       begin
-        frontmatter = ::YAML.load(Regexp.last_match[1])
-        frontmatter = frontmatter.respond_to?(:to_h) ? frontmatter.to_h : Hash.new(frontmatter)
-        @markup.metadata.merge!(frontmatter)
-      rescue Psych::SyntaxError
+        frontmatter = ::YAML.safe_load(@markup.sanitize.clean(Regexp.last_match[1]))
+        @markup.metadata.merge!(frontmatter) if frontmatter.respond_to?(:keys) && frontmatter.respond_to?(:values)
+      rescue ::Psych::SyntaxError => e
```
bartkamphorst: I prefer `error` over `e` for variable name.

lib/gollum-lib/page.rb
```diff
@@ -172,6 +172,13 @@ def metadata_title
       nil
     end
 
+    # Public: Whether or not to display the metadata
+    def display_metadata?
+      return false unless !metadata.empty? && !(metadata.size == 1 && metadata['title'])
```
bartkamphorst: I think the following is equivalent but better expresses the intention:
```ruby
return false if metadata.empty? || (metadata.size == 1 && metadata['title'])
```

bartkamphorst:
Final comments submitted, great stuff. 👍

PR #257, opened 2017-04-09
Allow same filename with different exts

dometto:
When playing around with https://github.com/gollum/gollum/pull/1220 I discovered that `gollum-lib` was still throwing `DuplicatePage` errors when creating e.g. `page.md` when `page.txt` already exits. This PR takes care of the issue, throwing `DuplicatePage` only when both file name and extension matches a file in the repo. It also removes the unused and confusing `allow_same_exts` argument from `Committer#add_to_index`.


bartkamphorst:
👍

PR #258, opened 2017-04-09
Refactor Page and Markup classes

dometto:


dometto:
* `url_path_display` is no longer necessary (after removal of ws subs it was identical to `url_path`)
* remove unnecessary `Page.canonicalize_filename` method
* change the `Markup.register` method so that it passes an array of valid file extensions rather than a regex
  * provide a `Markup.extensions` methods that returns all the registered page extensions
  * refactor `Page#valid_extension?` and `Page#format_for`
  * **NB**: This is because I assume `Markup.extensions.include?` is faster than iterating over `Markup.formats`, but I'm not sure. What do you think?

bartkamphorst:
Changing `Markup.register` is likely to break many people's custom configs. Could we leave a nice message for people calling `register` with a regexp to explain what they have to change in 5.x to get their `config.rb` working again?

dometto:
>  Could we leave a nice message for people calling register with a regexp to explain what they have to change in 5.x to get their config.rb working again?

What do you propose? Just a `$STDERR.puts "<warning>"`? That may still go unnoticed, but I think that's what I would prefer. Any more drastic measures like throwing an exception seem unnecessary, since the valid extensions for a markup default to `[ext_to_s]`, which should work in most cases.

lib/gollum-lib/markup.rb
```diff
@@ -43,10 +48,12 @@ def formats
       # If given a block, that block will be registered with GitHub::Markup to
       # render any matching pages
       def register(ext, name, options = {}, &block)
+        new_extension = options.fetch(:extensions, [ext.to_s])
         @formats[ext] = { :name => name,
-          :regexp => options.fetch(:regexp, Regexp.new(ext.to_s)),
+          :extensions => new_extension,
           :reverse_links => options.fetch(:reverse_links, false),
           :enabled => options.fetch(:enabled, true) }
+        @extensions.concat new_extension
```
bartkamphorst: I would like parentheses around the `concat` method (or else use `<<`).

bartkamphorst:
My bad, I was under the impression that the method signature had changed. I now see that in `5.x` regexps will still be accepted, just silently ignored. A warning to `sttderr` as you suggest might still be a nice touch though if `options[:regexp]`.

dometto:
No problem, it's good that we're now throwing the warning. I'll also add something on this in the release notes. PR should be finished and passing now.

bartkamphorst:
👍

PR #260, opened 2017-04-12
Implement Page#sha

dometto:


lib/gollum-lib/page.rb
```diff
@@ -88,6 +88,13 @@ def initialize(wiki)
       @parent_page    = nil
     end
 
+    # Public: The SHA hash identifying this page
+    #
+    # Returns the String SHA.
+    def sha
+      @blob && @blob.id
```
bartkamphorst: A little peculiar, this. Works though. 😄

dometto: Yeah, I noticed this pattern being used elsewhere in `page.rb`, too, so I decided to stick with the style. We could refactor it throughout the codebase if we don't like it. :)

bartkamphorst: It's a fun little rubyism, I don't mind. :)

PR #261, opened 2017-04-12
Proposal to remove WSD and switch to PlantUML by default. Fixes #250

### LINKED ISSUE #250: Remove WSD support from 5.x
I propose we remove support for Web Sequence Diagrams from `5.x`. WSD is a commercial platforms, and we currently have support for PlantUML already. Keeping multiple UML renderers will increase the maintenance load.

dometto:
As per #250. We can switch to using PlantUML only, since there is also a public server at http://plantuml.com (end users thus do not *have* to run their own server, but they can). One potential issue with making that the default is that existing PlantUML users who used the default server at `localhost` will silently switch to using the public server when upgrading to `5.x`.


bartkamphorst:
+1 for removing WSD Filter. 

> One potential issue with making that the default is that existing PlantUML users who used the default server at localhost will silently switch to using the public server when upgrading to 5.x.

So why the change of default server?

dometto:
> So why the change of default server?

The pro would be that there would be UML support out of the box. Not sure if that weighs up against the con.

dometto:
Ping @bartkamphorst. Shall we change the default server, or leave things as is?

bartkamphorst:
You're probably right that having it work out of the box is preferable for most users. Those who want to use their own server still can. Just make sure to add the change in default server to the changelog. 👍

dometto:
I've added the info to `HISTORY.md`, and have also started [this section](https://github.com/gollum/gollum/wiki/5.0-release-notes#migrating-your-wiki) in the release notes in the wiki. We should list changes that require changes to existing wikis there.

PR #262, opened 2017-04-12
Fix page url title

dometto:
[Previous refactor](https://github.com/gollum/gollum-lib/pull/258) had the result that page titles included their file extensions, thus breaking the `gollum` test. This PR fixes this issue.


dometto:
Looks like the WSD tests are failing on Travis due to problems with the service. Just a couple of hours after I opened [a PR](https://github.com/gollum/gollum-lib/pull/261) to remove WSD from gollum. Coincidence? I think not.

Restarting the Travis jobs...

PR #263, opened 2017-04-14
Fix anchors in links to absent pages.

dometto:
See https://github.com/gollum/gollum/issues/1096

### DIFF
--- a/lib/gollum-lib/filter/tags.rb
+++ b/lib/gollum-lib/filter/tags.rb
@@ -103,7 +103,7 @@ class Gollum::Filter::Tags < Gollum::Filter
     resolved_page_name = ::File.expand_path(page_name, "/"+@markup.dir)
 
     if @markup.include_levels > 0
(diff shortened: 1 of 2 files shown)


dometto:
A new edge case I just thought of: an anchored link to a page named `Potato#1.md`. It will first look for the page `Potato#1.md#anchor`, and when that fails look for `Potato` and treat `#1.md#anchor` as the anchor.

Can/should we fix this by looking for the *last* rather than the *first* instance of `#` and treat that as the anchor? That would break anchored links in which the anchors itself contains the `#` character, e.g. `Potato#achor#doubleanchor`, but I'm not sure if that's valid HTML anyway.

dometto:
Anchors containing `#` seem to be illegal, see e.g. [here](http://stackoverflow.com/questions/13570179/url-with-multiple-anchor-tag)

dometto:
This should fix things. Have also added a couple of tests to deal with further edge cases.

PR #266, opened 2017-04-20
Update sanitization to allow MathML

dometto:
See https://github.com/gollum/gollum/issues/1079

With the help of the Elements and Attributes reference for MathML mentioned there by @bartkamphorst, and some mean scripting described in [this gist](https://gist.github.com/dometto/52d9cb8b45d68bfc7665e5e6683b75a0), I managed to hark together a complete sanitize `Array` and `Hash` for MathML elements and attributes, respectively.

Shall we merge this into `master` (current PR) or into `5.x`? Not sure if this should be considered a breaking change.

### DIFF
--- a/gemspec.rb
+++ b/gemspec.rb
@@ -8,7 +8,7 @@ def specification(version, default_adapter, platform = nil)
     s.name              = 'gollum-lib'
     s.version           = version
     s.platform          = platform if platform
-    s.date              = '2017-04-04'
+    s.date              = '2017-04-08'
     s.rubyforge_project = 'gollum-lib'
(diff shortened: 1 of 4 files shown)


bartkamphorst:
Haha, nicely done! I think we should push it to 5.x.

dometto:
Turns out `Hash#transform_values` is a `ruby > 2.4` method only. So instead of modifying the normal `MATHML_ATTRS` hash by adding the `MATHML_COMMON_ATTRS` to each element in the hash, I'll now just make the script print the `MATHML_ATTRS` hash allong with the common attrs.

PR #269, opened 2017-04-23
Refactor tags

dometto:
This refactor shaves about 5 seconds of load time for a page with a ridiculous amount (1000) of tags (~25 seconds vs. ~30), and (hopefully) improves readability and maintainability. To further improve performance I would like to eventually collapse the search for files and pages so we don't traverse the repository twice, and of course implement caching of the repo tree map. But that requires refactoring the `Page` and `File` classes, and implementing the idea for a `Gollum::IO` class...

Some breaking changes:
  * syntax for link attributes changed from `image.jpg|frame|alt=Bla` to `image.jpg|frame, alt=Bla`
  * image links and file links and are now also reversed if `@markup.reverse_links?` (previously only for page links and external links). This is because [reversal is now handled at the start of the link processing rather than in each method separately](https://github.com/gollum/gollum-lib/pull/269/files#diff-9c66db1000504e8fd86544f23961803eR100).
  * a page link to `waa.md` would *first* be interpreted as a relative link, and if that didn't find a page, it would be interpreted as an absolute link (to `/waa.md`). So in cases of broken links the repository was traversed twice. I've disabled this, so an absolute link now requires an explicit `/`. However, this change doesn't break any tests.

### DIFF
--- a/lib/gollum-lib/filter/tags.rb
+++ b/lib/gollum-lib/filter/tags.rb
@@ -55,6 +55,7 @@ class Gollum::Filter::Tags < Gollum::Filter
   private
 
   PREFORMATTED_TAGS = %w(code tt)
+  INCLUDE_TAG = 'include:'
 
   def register_tag(tag)
     id       = "TAG#{Digest::SHA1.hexdigest(tag)}TAG"
@@ -75,42 +76,54 @@ class Gollum::Filter::Tags < Gollum::Filter
   # Returns the String HTML version of the tag.
   def process_tag(tag)
     link_part, extra = parse_tag_parts(tag)
-    mime = MIME::Types.type_for(::File.extname(link_part.to_s)).first
+    return generate_link('', nil, nil, :page_absent) if link_part.nil?
+    img_args = extra ? [extra, link_part] : [link_part]
+    mime = MIME::Types.type_for(::File.extname(img_args.first.to_s)).first
 
-    result = if link_part =~ /^_TOC_/
+    result = if tag =~ /^_TOC_/
       %{[[#{tag}]]}
     elsif link_part =~ /^_$/
       %{<div class="clearfloats"></div>}
-    elsif link_part =~ /^include:/
+    elsif link_part =~ /^#{INCLUDE_TAG}/
       process_include_tag(link_part)
     elsif mime && mime.content_type =~ /^image/
-      process_image_tag(link_part, extra)
-    elsif external = process_external_link_tag(extra, link_part)
+      process_image_tag(*img_args)
+    elsif external = process_external_link_tag(link_part, extra)
       external
     end
     result ? result : process_link_tag(link_part, extra)
   end
 
-  def process_link_tag(link_part, extra)
-    process_file_link_tag(link_part, extra) || process_page_link_tag(link_part, extra)
+  # Process the tag parts as an internal link to a File or Page. 
+  def process_link_tag(link_part, pretty_name)
+    process_file_link_tag(link_part, pretty_name) || process_page_link_tag(link_part, pretty_name)
   end
-
+  
+  # Parse the tag (stuff between the double brackets) into a link part and additional information (a pretty name, description, or image options).
+  #
+  # tag       - The String tag contents (the stuff inside the double
+  #             brackets).
+  #
+  # Returns an Array of the form [link_part, extra], where both elements are Strings and the second element may be nil.
   def parse_tag_parts(tag)
-    parts = tag.split('|').map(&:strip)
+    parts = tag.split('|').map(&:strip)[0..1]
     parts.reverse! if @markup.reverse_links?
-    parts
+    if parts[1]
+      return parts[1], parts[0]
+    else
+      return parts[0], nil
+    end
   end
 
   # Attempt to process the tag as an include tag
   #
   # tag - The String tag contents (the  stuff inside the double brackets).
   #
-  # Returns the String HTML if the tag is a valid image tag or nil
-  #   if it is not.
-  #
+  # Returns the String HTML if the tag includes a valid page or an error message if the page could not be found.
   def process_include_tag(tag)
-    return html_error('Cannot process include directive: no page name given') if tag.length <= 8
-    page_name          = tag[8..-1]
+    len = INCLUDE_TAG.length
+    return html_error('Cannot process include directive: no page name given') if tag.length <= len
+    page_name          = tag[len..-1]
     resolved_page_name = ::File.expand_path(page_name, "#{::File::SEPARATOR}#{@markup.dir}")
     if @markup.include_levels > 0
       page = find_page_from_path(resolved_page_name)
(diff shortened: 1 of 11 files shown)


dometto:
Todo: update inline docs for tag filter.

lib/gollum-lib/filter/tags.rb
```diff
@@ -4,7 +4,8 @@
 class Gollum::Filter::Tags < Gollum::Filter
   # Extract all tags into the tagmap and replace with placeholders.
   def extract(data)
-    return data if @markup.format == :txt || @markup.format == :asciidoc
+    return data if @markup.skip_tags?
```
bartkamphorst: Expresses the intent much clearer, nice. 👍

lib/gollum-lib/filter/tags.rb
```diff
@@ -73,21 +74,31 @@ def is_preformatted?(node)
...
+    elsif mime && mime.content_type =~ /^image/
+      process_image_tag(link_part, extra)
+    elsif external = process_external_link_tag(extra, link_part)
+      external
     end
+    result ? result : process_link_tag(link_part, extra)
+  end
+
+  def process_link_tag(link_part, extra)
+    process_file_link_tag(link_part, extra) || process_page_link_tag(link_part, extra)
```
bartkamphorst: This means that all tags are processed as file links initially. Is there any way to avoid this?

lib/gollum-lib/filter/tags.rb
```diff
@@ -98,10 +109,9 @@ def process_tag(tag)
   #   if it is not.
   #
   def process_include_tag(tag)
-    return unless /^include:/.match(tag)
+    return html_error('Cannot process include directive: no page name given') if tag.length <= 8
```
bartkamphorst: What is this magic number 8?

lib/gollum-lib/filter/tags.rb
```diff
@@ -98,10 +109,9 @@ def process_tag(tag)
...
     page_name          = tag[8..-1]
-    resolved_page_name = ::File.expand_path(page_name, "/"+@markup.dir)
-
+    resolved_page_name = ::File.expand_path(page_name, "#{::File::SEPARATOR}#{@markup.dir}")
```
bartkamphorst: Nice (helpful going forward on Windows). I guess it doesn't make sense to work with Pathname objects here, right?

lib/gollum-lib/filter/tags.rb
```diff
@@ -120,23 +130,19 @@ def process_include_tag(tag)
...
-    name = parts[0].strip
-    if (file = @markup.find_file(name))
-      path = ::File.join @markup.wiki.base_path, file.path
-    elsif name =~ /^https?:\/\/.+(jpg|png|gif|svg|bmp)$/i
+  def process_image_tag(name, options = nil)
+    if name =~ /^https?:\/\/.+$/i
       path = name
-    elsif name =~ /.+(jpg|png|gif|svg|bmp)$/i
+    elsif (file = @markup.find_file(name))
+      path = ::File.join @markup.wiki.base_path, file.path
```
bartkamphorst: Stylistically I'd like to see the join call have parentheses.

lib/gollum-lib/filter/tags.rb
```diff
@@ -246,12 +246,7 @@ def process_external_link_tag(tag)
...
   #   if it is not.
-  def process_file_link_tag(tag)
-    parts = tag.split('|')
-    return if parts.size.zero?
-
-    name = parts[0].strip
-    path = parts[1] && parts[1].strip
+  def process_file_link_tag(name, path)
     if path && file = @markup.find_file(path)
       path = ::File.join @markup.wiki.base_path, file.path
```
bartkamphorst: Parentheses.

bartkamphorst: This methods returns HTML. Is there any way to refactor that so that gollum (and not gollum-lib) wraps `path` and `name` in the appropriate html tags?

lib/gollum-lib/filter/tags.rb
dometto: Yes, that's what bothered me originally. I would like to just walk through the repo searching for the requested paths, returning `Page` objects if the found object has a valid page extensions and `File` objects otherwise. Then `process_link_tag` should choose what to with each link based on what kind of object it gets back (if any).

dometto: But I think that may require a new approach to `Page` and `File` objects altogether. Perhaps have `BlobEntry` superclass of which `Page` and `File` are both subclasses, with a common find-and-construct-from-path method. Or just have the `BlobEntry` class with a `#file?` and `#page?` method?

lib/gollum-lib/filter/tags.rb
dometto: :) Point take 👍

lib/gollum-lib/filter/tags.rb
dometto: Not sure. I tried putting a `::File.join` here but that broke some tests. The point here is to prevent people from including `/etc/password`. `File.expand_path` here never returns a path lower in the directory tree than `@markup.dir`. Does `Pathname` have a similar method?

bartkamphorst: I'm not sure, but we could just build a Pathname object from the string that `expand_path` comes up with. The bigger question is whether we need something more than a string here. I guess time will tell. ;)

dometto: Right. If we want to rebuild the API for searching/retrieving objects from the repo to use `Pathname`, we should also do it here.

dometto:
@bartkamphorst: incorporated your stylistic suggestions, @bartkamphorst, and refactored further.

I don't see how we can easily get rid of returning HTML inside the tag filter, since the purpose of that filter (and other filters) *is* basically to generate HTML on top of what the markup parser generates. However, I've tried to at least cleanly separate the tag-parsing logic from the html-generation logic. Hardcoded HTML is now only present in `#generate_link` and `#generate_image` and one or two helper methods they call.

One way to get the hardcoded HTML out of `gollum-lib` would perhaps be to move `#generate_link` and `#generate_image` to `gollum`, and somehow pass them on to the tag filter as a `Proc` from there on? But that would mean that the filter by itself would not be fully functional 'out of the box'.

dometto:
And as you suggested, `process_file_link_tag` and `process_page_link_tag` should be collapsed into a single method. But I would propose to merge this now, and to take that further step once we have refactored the `Page` and `File` classes.

dometto:
Another change in behavior affected by this PR is that `File` links without a description now work, whereas in `master` they seem to result in a broken `img` for some reason. The latter clearly seems like a bug, so I'll add a regression test.

dometto:
Re-implemented the tag skipping logic in the `Markup` class as you suggested, reverted the changes to the `Filter` class. I think this is nice and clean now, thanks for the review!

lib/gollum-lib/markup.rb
```diff
@@ -126,7 +126,7 @@ def render_default(data, format=:markdown, name='render_default.md')
       @format = format
       @name   = name
 
-      chain = [:YAML, :PlainText, :Emoji, :TOC, :RemoteCode, :Code, :Sanitize, :PlantUML, :Tags, :Render]
+      chain = [:YAML, :PlainText, :Emoji, :TOC, :RemoteCode, :Code, :Sanitize, :PlantUML, :Tags, :Render].reject {|filter| skip_filter?(filter)}
```
bartkamphorst: Why here?

dometto: You mean why in the `render_default` method? It's used in the tests at least, so filter skipping should work there too, I thought.

bartkamphorst: Because you also call `.reject` on line 180.

dometto: Right, but the `render` and `render_default` methods are enitrely separate from each other, so the rejecting needs to happen twice. But I now notice that `render_default` is called just once in `gollum-lib` and `gollum` taken together, namely in a test. If I modify that test we can just remove this method alltogeter. :+1:

bartkamphorst: 👍

dometto: `render_default` was being used to allow use of the `Markup` class without a wiki. I can see some usecases for that, but I don't know we should consider them within the legitimate scope of the gollum project. Perhaps in the interest of general code sanity, we can remove it?

bartkamphorst: Go for it. 😄 People can always come knocking if they miss this functionality in 5.x.

bartkamphorst: Might be good to add to the 5.x changelog though!

PR #270, opened 2017-04-29
Refactor Markup class

dometto:
* Moved the cleanup of empty `<p></p>` tags to the `Sanitize` filter, which was actually causing them to be introduced.
* Removed mysterious line

Todo:
* what's with the empty caching methods, `#update_cache` and `#check_cache`?
* can we make the complicated `class << self` block at the top more readable? I find it a bit of an odd way to handle the definition of class methods.
* The `#find_files` method should be removed, but this will require some modifications to the tag filter.

Any further suggestions welcome!

### DIFF
--- a/lib/gollum-lib/filter/sanitize.rb
+++ b/lib/gollum-lib/filter/sanitize.rb
@@ -10,7 +10,7 @@ class Gollum::Filter::Sanitize < Gollum::Filter
       doc = Nokogiri::HTML::DocumentFragment.parse(data)
       doc = @markup.sanitize.clean_node!(doc)
 
-      doc.to_xml(@markup.to_xml_opts)
+      doc.to_xml(@markup.to_xml_opts).gsub(/<p><\/p>/, '')
     else
       data
     end
--- a/lib/gollum-lib/markup.rb
+++ b/lib/gollum-lib/markup.rb
@@ -129,11 +129,6 @@ module Gollum
         data = filter.process(data)
       end
 
-      # Finally, a little bit of cleanup, just because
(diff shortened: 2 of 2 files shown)


bartkamphorst:
> what's with the empty caching methods, #update_cache and #check_cache?

It defines an interface. It's called in the code filter. Was probably implemented elsewhere in Github's production code to connect it to a nosql store or something. I'd keep it in.

bartkamphorst:
> Removed mysterious line

There used to be [MarkupGFM class](https://gitlab.com/gitlab-org/gollum-lib/commit/0cc4c9f2c299ed8dc75d13fbe3c7b67f3fb7940c) (a subclass of `Markup`), and this line was [kept for compatibility](https://gitlab.com/gitlab-org/gollum-lib/commit/b5a2a8df67e7b5b74469d5a55b4e33cc745061e9). I think it is okay to remove it in 5.x. 👍

bartkamphorst:
> can we make the complicated class << self block at the top more readable? I find it a bit of an odd way to handle the definition of class methods.

That's actually very [idiomatic in Ruby](http://www.integralist.co.uk/posts/eigenclass.html), and I honestly prefer it to explicitly defining all class methods with `def self.methodname`.

bartkamphorst:
> The #find_files method should be removed, but this will require some modifications to the tag filter.

Agreed. 👍

dometto:
> That's actually very idiomatic in Ruby, and I honestly prefer it to explicitly defining all class methods with def self.methodname.

I stand corrected. :)

PR #271, opened 2017-04-29
Render footnotes and citations via pandoc

dometto:
Attempt to implement support for `pandoc` by default, without causing conflicts with other markup renderers. Added a test for the new filter, and everything seems fine.

Note: I removed support for BibTeX pages from this PR, since that's separate functionality that should have its own PR.

### DIFF
--- a/lib/gollum-lib/file.rb
+++ b/lib/gollum-lib/file.rb
@@ -26,6 +26,13 @@ module Gollum
       path
     end
 
+    # Public: The SHA hash identifying this file
+    #
+    # Returns the String SHA.
+    def sha
+      @blob && @blob.id
+    end
+
     # Public: The url_path, but CGI escaped.
     #
     # Returns the String url_path
--- a/lib/gollum-lib/filter/pandoc_bib.rb
+++ b/lib/gollum-lib/filter/pandoc_bib.rb
@@ -4,7 +4,7 @@ require 'pathname'
 # When using pandoc, put relevant bibliography metadata extracted in the YAML filter back in the document so it gets passed on to pandoc.
 class Gollum::Filter::PandocBib < Gollum::Filter
 
-  BIB_KEYS = ['bibliography', 'csl']
+  BIB_KEYS = ['bibliography', 'csl', 'link-citations', 'nocite']
   
   def process(data)
     data
--- a/lib/gollum-lib/filter/pandoc_bib.rb
+++ b/lib/gollum-lib/filter/pandoc_bib.rb
@@ -1,0 +1,39 @@
+require 'yaml'
+require 'pathname'
+
+class Gollum::Filter::PandocBib < Gollum::Filter
+
+  BIB_KEYS = ['bibliography', 'csl']
+  
+  def process(data)
+    data
+  end
+
+  def extract(data)
+    return data unless @markup.format == :markdown && @markup.metadata && bibliography_metadata_present? && using_pandoc?
+    bib_metadata = {}
+
+    BIB_KEYS.each do |bibliography_key|
+      path = @markup.metadata[bibliography_key]
+      next unless file = @markup.wiki.file(path)
+      path = Pathname.new("#{::File.join(::Dir.tmpdir, file.sha)}#{::File.extname(path)}")
+      bib_metadata[bibliography_key] = path.to_s
+      unless path.exist?
+        path.open('w') do |copy_file|
+          copy_file.write file.raw_data
+        end
+      end
+    end
+    bib_metadata.empty? ? data : "#{bib_metadata.to_yaml}---\n#{data}"
+  end
+
+  private
+
+  def using_pandoc?
+    GitHub::Markup::Markdown.new.implementation_name == 'pandoc-ruby'
+  end
+
+  def bibliography_metadata_present?
+    @markup.metadata.keys.any? {|key| BIB_KEYS.include?(key)}
+  end
+end
--- a/lib/gollum-lib/filter/pandoc_bib.rb
+++ b/lib/gollum-lib/filter/pandoc_bib.rb
@@ -1,6 +1,7 @@
 require 'yaml'
 require 'pathname'
 
+# When using pandoc, put relevant bibliography metadata extracted in the YAML filter back in the document so it gets passed on to pandoc.
 class Gollum::Filter::PandocBib < Gollum::Filter
 
   BIB_KEYS = ['bibliography', 'csl']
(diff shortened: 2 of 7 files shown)


dometto:
An even stronger implementation would of course allow using pandoc to render any of the formats it supports.. :)

lib/gollum-lib/markups.rb
```diff
@@ -28,11 +28,24 @@ def executable_exists?(name)
...
+  end
+end
+
 module Gollum
   class Markup
     GitHub::Markup::Markdown::MARKDOWN_GEMS['kramdown'] = proc { |content|
         Kramdown::Document.new(content, :auto_ids => false, :smart_quotes => ["'", "'", '"', '"'].map{|char| char.codepoints.first}).to_html
     }
+    GitHub::Markup::Markdown::MARKDOWN_GEMS['pandoc-ruby'] = proc { |content|
+        PandocRuby.convert(content, :s, :from => :markdown, :to => :html, :filter => 'pandoc-citeproc')
```
dometto: Possibly make this conditional on the availability of the pandoc-ruby gem? If it isn't available it shouldn't matter to github-markup though.

lib/gollum-lib/file.rb
```diff
@@ -26,6 +26,13 @@ def url_path
       path
     end
 
+    # Public: The SHA hash identifying this file
+    #
+    # Returns the String SHA.
+    def sha
+      @blob && @blob.id
```
bartkamphorst: Is `@blob` ever nil for a Gollum::File? I suppose they might be if we start serving uncommitted files, right? Otherwise the `@blob` check is superfluous.

lib/gollum-lib/filter/pandoc_bib.rb
```diff
@@ -0,0 +1,39 @@
...
+    bib_metadata = {}
+
+    BIB_KEYS.each do |bibliography_key|
+      path = @markup.metadata[bibliography_key]
+      next unless file = @markup.wiki.file(path)
+      path = Pathname.new("#{::File.join(::Dir.tmpdir, file.sha)}#{::File.extname(path)}")
+      bib_metadata[bibliography_key] = path.to_s
+      unless path.exist?
+        path.open('w') do |copy_file|
+          copy_file.write file.raw_data
```
bartkamphorst: Parentheses around write call.

test/filter/test_emoji.rb
```diff
@@ -1,7 +1,6 @@
 # ~*~ encoding: utf-8 ~*~
 path = File.join(File.dirname(__FILE__), "..", "helper")
 require File.expand_path(path)
-require 'ostruct'
```
bartkamphorst: Does this belong in this PR? I don't really mind, just wondering about its significance. 😄

test/helper.rb
```diff
@@ -65,15 +65,24 @@ def commit_details
     :email   => "tom@github.com" }
 end
 
+class MockWiki
+  def file(path)
+    OpenStruct.new(
+      :sha => 'dasd',
```
bartkamphorst: Use "actual" fake sha instead?

lib/gollum-lib/markups.rb
bartkamphorst: Github-markup's behavior might change, though. To be on the safe side, it's probably wise to check for gem availability ourselves.

lib/gollum-lib/filter/pandoc_bib.rb
```diff
@@ -0,0 +1,39 @@
...
+        end
+      end
+    end
+    bib_metadata.empty? ? data : "#{bib_metadata.to_yaml}---\n#{data}"
+  end
+
+  private
+
+  def using_pandoc?
+    GitHub::Markup::Markdown.new.implementation_name == 'pandoc-ruby'
```
bartkamphorst: Not sure I understand why the Markdown class needs to be instantiated here.

lib/gollum-lib/file.rb
dometto: It can be `nil` because `File`s are first initialized, and then populated in a second step. This strikes me as an ugly and unnecessary pattern, which we can hopefully change when we rework the `File` and `Page` classes.

test/filter/test_emoji.rb
dometto: I first started what evolved into the `MockWiki` `mock_page` helpers in this file in a previous PR, but they ended up in `helpers.rb`. Just forgot to remove the `require` here. :)

bartkamphorst: 👍 Carry on. 😉

lib/gollum-lib/file.rb
bartkamphorst: OK, thanks for clarifying that for me.

lib/gollum-lib/filter/pandoc_bib.rb
dometto: I wanted to implement `implementation_name` as a class method, but it uses the `GitHub::Markup::Markdown#try_require` method, which is a private instance method. Perhaps we can use `instance_eval` to get around the problem, or just create a new instance inside the class method.

bartkamphorst: Don't we have an instantiated Markdown object stored somewhere already? Otherwise I prefer creating a new instance inside the class method to using `instance_eval`.

dometto:
Implemented your suggestions. Thanks again! 👍

dometto:
@bartkamphorst, do you think this is ready to merge? Also understand if you think it's too hacky (given the overrides and the need for an extra filter).

dometto:
Refactored so that, if the wiki is non-bare, pandoc will simply be passed the path of the relevant bibfiles in the repo (rather than copying its contents to a tempfile).

bartkamphorst:
:+1:

bartkamphorst:
I think this is an excellent addition in terms of functionality, and I think you've implemented it clearly and cleanly. With regard to the additional Filter though, couldn't we remove the Filter from the filter_chain if we know that it won't be used (either because it is disabled, or because certain requirements are not met)?

dometto:
Thanks!

> couldn't we remove the Filter from the filter_chain if we know that it won't be used (either because it is disabled, or because certain requirements are not met)?

Could do that [here](https://github.com/gollum/gollum-lib/blob/master/lib/gollum-lib/wiki.rb#L252-L254)?

bartkamphorst:
> Could do that here?

Exactly, comparable to how it is done with Emoji filter.

dometto:
Something like this then? :)

bartkamphorst:
Yes, makes more sense not to run through any filter when it is not being used.

dometto:
All lights green 👍

bernatguillen:
I hope this is the right place to ask: Is there a way to make a hacky version of this within a config.rb file (i.e. basically copying pandoc_bib.rb and adding the filter) or should I wait for gollum 5?

dometto:
If you want it right away I think your best bet is to cherry pick this commit and merge it into your version of gollum. The PR is fairly self contained so you should be able to get it to work. No warrantee and no support though. :)

PR #272, opened 2017-04-30
Implement support for BibTeX pages.

dometto:
BibTeX support without clashing with other markups renderers. Allows setting of style to use via YAML frontmatter. It requires some extra dependencies, but they all look lightweight, well-maintained, and fortunately jruby-compatibile.

Note: if we merge this, we have access to a bunch of citation style files from `csl-styles`. We could even update the pandoc filter (#271) to use files from that gem, instead of using user-committed style files.

### DIFF
--- a/lib/gollum-lib/filter/bibtex.rb
+++ b/lib/gollum-lib/filter/bibtex.rb
@@ -1,6 +1,7 @@
 begin
   require 'bibtex'
   require 'citeproc'
+  require 'csl'
   require 'csl/styles'
 rescue LoadError => error
   puts "Error trying to require an optional gem for BibTeX parsing: #{error.to_s}"
@@ -11,15 +12,20 @@ class Gollum::Filter::BibTeX < Gollum::Filter
 
   def extract(data)
     return data unless supported_format? && gems_available? && bib = ::BibTeX.parse(data).convert(:latex)
-    style = @markup.metadata['bibstyle'] if @markup.metadata
+    style = find_csl_data('csl') || ::CSL::Style.default
+    locale = find_csl_data('locale') || ::CSL::Locale.default
+
     begin
-      style = ::CSL::Style.load(style.to_sym) if style
-    rescue ::CSL::ParseError
-      style = nil
+      style = ::CSL::Style.load(style)
+      ::CSL::Locale.load(locale)
+    rescue ::CSL::ParseError => error
+      log_failure(error.to_s)
+      return CGI.escapeHTML(data)
     end
-    citeproc = ::CiteProc::Processor.new(style: style || 'apa', format: 'html')
+
+    citeproc = ::CiteProc::Processor.new(style: style, locale: locale, format: 'html')
     citeproc.import(bib.to_citeproc)
-    citeproc.bibliography.references.join("<br/>")
+    citeproc.bibliography.references.join('<br/>')
   end
 
   def process(data)
@@ -28,11 +34,23 @@ class Gollum::Filter::BibTeX < Gollum::Filter
 
   private
 
+  def log_failure(msg)
+    @markup.metadata = {} unless @markup.metadata
+    @markup.metadata['errors'] = [] unless @markup.metadata['errors']
+    @markup.metadata['errors'] << "Could not render the bibliography because no valid CSL or locale file was found in the wiki or in the CSL directory. Please commited a valid file, or install the csl-styles gem. The message from the parser was: #{msg.to_s}."
+  end
+
   def supported_format?
     @markup.format == :bib
   end
 
   def gems_available?
-    ::Gollum::MarkupRegisterUtils::gems_exist?(["bibtex-ruby", "citeproc-ruby", "csl"])
+    ::Gollum::MarkupRegisterUtils::gems_exist?(['bibtex-ruby', 'citeproc-ruby', 'csl'])
+  end
+
+  def find_csl_data(key)
+    path = @markup.metadata ? @markup.metadata[key] : nil
+    file = path ? @markup.wiki.file(path) : nil
+    file.nil? ? path : file.raw_data
   end
 end
--- a/lib/gollum-lib/filter/bibtex.rb
+++ b/lib/gollum-lib/filter/bibtex.rb
@@ -1,0 +1,38 @@
+begin
+  require 'bibtex'
+  require 'citeproc'
+  require 'csl/styles'
+rescue LoadError => error
+  puts "Error trying to require an optional gem for BibTeX parsing: #{error.to_s}"
+end
+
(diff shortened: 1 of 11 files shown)


dometto:
...oops:
```
$ bundle show csl-styles
~/.rvm/gems/ruby-2.4.1@gollum-lib/gems/csl-styles-1.0.1.7
$ du -sh ~/.rvm/gems/ruby-2.4.1@gollum-lib/gems/csl-styles-1.0.1.7
 41M	~/.rvm/gems/ruby-2.4.1@gollum-lib/gems/csl-styles-1.0.1.7
```

dometto:
Because of the large size of the `csl-styles` gem I've now updated the filter and tests to run without it, although it will take advantage of it when present. The filter will now try to utilize csl and locale files in this order:

* if the `csl` or `locale` key is specified in the frontmatter, it will first try to load files from the repository at the locatation specified by the metadata `value`
* if that fails, it will attempt to load a file named `value` from `CSL::Style.root` and `CSL::Locale.root` respectively.
* if no locale or csl file was specified, it will try to load `CSL::Style.default` and `CSL::Locale.default`.
* if all this fails, rendering will fail and an error message will be appended to `metadata['errors']`.

I've added some example files to the test directory in order to get the tests off the ground without the gem present. The tests pass, and I can also run `gollum` with this branch and get l33t BibTeX rendering. :)

lib/gollum-lib/filter/bibtex.rb
```diff
@@ -0,0 +1,56 @@
+begin
+  require 'bibtex'
+  require 'citeproc'
+  require 'csl'
+  require 'csl/styles'
+rescue LoadError => error
+  puts "Error trying to require an optional gem for BibTeX parsing: #{error.to_s}"
```
bartkamphorst: Not sure if this should stay in. If it does, it should write to `STDERR`.

lib/gollum-lib/markups.rb
```diff
@@ -10,6 +9,13 @@ def gem_exists?(name)
       Gem::Specification.find {|spec| spec.name == name} != nil
     end
 
+    def gems_exist?(names)
```
bartkamphorst: Though this method behaves idiomatically (question mark indicating a boolean return value), based on its name, I would actually expect this method to return an array of booleans (one for each gem). Instead, it returns false if a single gem does not exist. I would prefer to rename this method to `all_gems_available?`.

lib/gollum-lib/markups.rb
```diff
@@ -59,5 +65,8 @@ class Markup
              :extensions => ['mediawiki','wiki'], :reverse_links => true)
     register(:pod, "Pod",
              :enabled => MarkupRegisterUtils::executable_exists?("perl"))
+    register(:bib, "BibTeX", :extensions => ['bib'],
+             :enabled => MarkupRegisterUtils::gems_exist?(["bibtex-ruby", "citeproc-ruby", "csl-styles"]),
```
bartkamphorst: As a general remark: can end users easily figure out why `:enabled` would be false? For example, say that I have failed to install `citeproc-ruby`, how would I know that this is the case?

lib/gollum-lib/markups.rb
```diff
@@ -59,5 +65,8 @@ class Markup
...
+             :skip_filters => Proc.new {|filter| ![:YAML,:BibTeX,:Sanitize].include?(filter)})
```
bartkamphorst: This is a little cryptic for me.

dometto: The `skip_filters` `Proc` or the whole method call?

lib/gollum-lib/markups.rb
dometto: I think where this shows up for the end user is currently only the dropdown menu in the editor, where unavailable markups will be red (or grey or some such) and unselectable. I actually think it might be a nice idea to have something like a gollum info or settings page where this kind of stuff is also displayed.

lib/gollum-lib/filter/bibtex.rb
dometto: Yeah, I also wasn't sure. I always find it a bit strange to rescue an exception and just ignore it, but in this case I think we should maybe remove it: I keep seeing the error when running the tests. :)

lib/gollum-lib/markups.rb
dometto: Ah, your questions was *how* users know this. There is currently nothing saying that. Perhaps we should change this to an `Array` of `:required_gems` (and set the value `:enabled` inside the `register` method based on whether these gems are available). Then on the the future info page, we can list the gem requirements for each markup?

lib/gollum-lib/markups.rb
dometto: Never mind, github was being a bit ambiguous, but it's now clear that this refers to line 70. :)

What about: `Proc.new {|filter| true unless [:YAML,:BibTeX,:Sanitize].include?(filter)}`

test/examples/bibtex/LICENSE.md
```diff
@@ -0,0 +1,14 @@
+## License
+
+These CSL and locale files were copied form the csl-styles gem (https://github.com/inukshuk/csl-styles).
```
dometto: Note to self: typo

lib/gollum-lib/markups.rb
bartkamphorst: I really like the idea of having an info page, but the worry of course is that all users would be able to view it. Not sure how I feel about that.

lib/gollum-lib/markups.rb
bartkamphorst: The Proc.

lib/gollum-lib/markups.rb
dometto: Yeah, that's true. The alternative is to make it a `--info` switch or include it in the `--version` output.

dometto:
Updated

dometto:
@bartkamphorst, some `jgit` issues are causing the tests to fail. Any idea what could be causing this? We haven't recently updated `rjgit`, right?

bartkamphorst:
Issue is wrong java version; needs version 8.

dometto:
Thanks for the explanation. Now green!

RensDimmendaal:
How can I use this function as a user?
I'm sorry if this is not the right place to ask, and if so I would appreciate it if you could point me in the right direction.

dometto:
@RensDimmendaal as a user you just commit a bibtex file to the wiki and load in the browser. However, this currently works only on the `5.x` development branch, which I would advise against using in production.

RensDimmendaal:
Thanks that helps.

PR #274, opened 2017-04-30
Escape emojis

svoop:
screen shot 2017-04-30 at 16 32 00


svoop:
Oops, wrong base :smile:

PR #276, opened 2017-05-08
Refactor wiki.rb: use Gollum class names directly

dometto:
I attempted to untangle the the mess of `wiki.blah_class.new` calls. I think this simplifies the code quite a lot, without sacrificing a lot in terms of flexibility. To override anything an end user can just monkeypatch or subclass any of the provided classes. My guess is this is all an old compatibility layer, as some of the comments in the removed code also indicate.

Still, if anyone thinks this provides useful flexibility, we can keep it.

### DIFF
--- a/lib/gollum-lib/page.rb
+++ b/lib/gollum-lib/page.rb
@@ -2,8 +2,6 @@
 module Gollum
   class Page
     include Pagination
-
-    Wiki.page_class = self
     
     SUBPAGENAMES = [:header, :footer, :sidebar]
     
@@ -240,7 +238,7 @@ module Gollum
     #
     # Returns a Gollum::Markup instance.
     def markup
-      @markup ||= @wiki.markup_classes[format].new(self)
+      @markup ||= ::Gollum::Markup.new(self)
     end
 
     # Public: The current version of the page.
--- a/lib/gollum-lib/blob_entry.rb
+++ b/lib/gollum-lib/blob_entry.rb
@@ -48,7 +48,7 @@ module Gollum
     # Returns a Gollum::Page instance.
     def page(wiki, commit)
       blob         = self.blob(wiki.repo)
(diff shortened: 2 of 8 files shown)


dometto:
@bartkamphorst ok to merge?

bartkamphorst:
I can't really oversee the consequences for people making use of this. Then again, I don't know if anyone actually is making use of this. So I don't mind either way at the moment.

dometto:
@bartkamphorst any new thoughts on this?

bartkamphorst:
No, not really. Perhaps we can ask the folks over at gitlab if they use this functionality?

dometto:
>  Perhaps we can ask the folks over at gitlab if they use this functionality?

👍  @fjsanpedro @douwem can you let us know if this PR would cause trouble on your end? I find the `wiki.page_class.new` pattern somewhat unrubyish, so I'd prefer to remove it and let people subclass or monkeypatch if they want different functionality. But perhaps I'm overlooking something.

fjsanpedro:
@dometto I don't recall any subclassing. I think we rely on `Gollum` in that layer. 

So I think we will be fine with the change 😄 (and if it's for the better we can always change things in our code).

lib/gollum-lib/page.rb
```diff
@@ -240,7 +238,7 @@ def format
     #
     # Returns a Gollum::Markup instance.
     def markup
-      @markup ||= @wiki.markup_classes[format].new(self)
```
fjsanpedro: @dometto I've skimmed the PR a little bit, and I have one question. How the `format` thing is going to be applied to this new approach? I mean in the old approach we got the proper `Markup` class based on the format, now we get the generic class `Gollum::Markup`. I supposed that the class will act accordingly to the page format?

fjsanpedro: Nice then 👍. Everything looks cleaner now. Good work!

PR #277, opened 2017-05-15
Implemented the Video Gollum Macro.

atrost:
Added support for html5 video through a Gollum macro. Rspec included, also tested complete scenario from uploads to editing and displaying in a page with Gollum running as a standalone app and embedded into a rails app.


test/test_macros.rb
```diff
@@ -1,6 +1,7 @@
 # ~*~ encoding: utf-8 ~*~
 require File.expand_path('../helper', __FILE__)
 require File.expand_path('../wiki_factory', __FILE__)
+require 'pry'
```
dometto: This can go I think

lib/gollum-lib/macro/video.rb
```diff
@@ -0,0 +1,17 @@
+module Gollum
+  class Macro
+    class Video < Gollum::Macro
+      def render (fname)
+        begin
```
dometto: Do we need this elaborate way of determining whether the video file exists? If the video tag is passed a path to a non-existent file, what will happen? Won't the browser just handle it as a broken link?

dometto: Another worry is that the `File.open` check will actually fail for bare git repositories, i.e. repositories that don't have a working directory (but that might still contain the video files).

dometto:
Thanks for the PR, @atrost! Some suggestions inline in the code, and some more general questions here.

* If I read the code correctly the generated link to a video file is something like `/somevid.mp4`, which the browser will interpret as a link to `http://localhost:4567/somevid.mp4` (or wherever your `gollum` is install is hosted). So `somevid.mp4` will end up being served by `gollum`. Does that work out of the box, or does it require any tinkering to the `gollum` sinatra app (for instance, are committed video files served with the correct mimetype at present)?
* Does this support links to externally hosted video files (`https://blah.com/somevid.mp4`)? I think not, because of the `File.open` check. Do we want to support that?

lib/gollum-lib/macro/video.rb
dometto: Also, even if we need to keep the existence check on the filesystem, you should simply use `File.exists?(pname)`, which will simply return `false` if the file isn't on the filesystem. That way you don't need the `begin/rescue/return` flow!

(Sorry for the multiple comments)

test/test_macros.rb
```diff
@@ -126,4 +127,22 @@ def render(opts)
     @wiki.write_page("ListNamedArgsPage", :markdown, "<<ListNamedArgs(xyzzy=\"Foo\")>>", commit_details)
     assert_match(/@xyzzy = Foo@/, @wiki.pages[0].formatted_data)
   end
+
+  test "Video macro given a valid mp4 file displays an html5 video tag with a video type of mp4" do
+    file = "/assets/vid.mp4"
```
dometto: Rather than committing the test video files to the repo, it may be cleaner to use [stubs](https://github.com/freerange/mocha) here: `File.stubs(:exists?).with(testpath('examples', file)).returns(true)`. Also see the above comment about using `File.exists?`. But again: it would be even better if we don't need to check for file existence, and can just leave that to the browser.

dometto:
Ah, I just realized what a potential worry is with skipping the file existence check and leaving broken link handling to the browser: if we link to `/assets/somevid.mp4` and it doesn't exist, gollum will serve a `404` page, which may lead to trouble? But I would guess most browsers have intelligent ways of handling video tags with sources that turn out not to have the right mimetype or content?

dometto:
Also, the travis tests fail because of issues with older versions of `jruby`. This has been fixed in the `gollum-lib-5.x` branch. I would propose just changing the base branch of this PR to that branch, if you don't mind?

lib/gollum-lib/macro/video.rb
atrost: The code does both, checks if the video file exists and also detects the video's MIME type so users don't need to specify it explicitly in the macro. With the current version of the macro, it generates the correct video/#{type} based on the video file being linked to.  We could detect the type based on a file extension, but that won't be considered a reliable way of detecting file types. This is also the reason why I committed video test files to the repo; a test case to cover the automatic mime type detection part. Regarding a bare repo and when a file doesn't exist in the file system, we could add a `File.exists?(pname)` before trying to open, but then what should we use for the mime type?  I am open to any suggestions. Thanks

atrost: I personally never used bare repos. Can they actually be used in deployments to production?

dometto: Yes, `gollum` supports the use of bare repositories. This is essential for use cases where users can also push to the repo over `http`, for example.

dometto: > detects the video's MIME type so users don't need to specify it explicitly in the macro.

Thanks for the explanation, I understand the need now. What about the following approach then?
* Change the macro to accept two arguments: `fname` and `mime` (optional)
* If the repo is bare (`@wiki.repo.is_bare?`, I think), and `mime` is not specified, throw an error.
* If the repo is not bare, and `mime` is not specified, check the mime type in the currently implemented way.
* If the repo is not bare, and `mime` is specified, use `mime` as mime type.

Technically it would be possible to do the mime type check on bare repositories as well, using the `gollum-lib` API as I explained above. But I think it would be better not to do this for performance reasons (for large videos, this would be quite an expensive operation).

dometto: Actually, according to [this](http://stackoverflow.com/a/7537309) the `MIME::Types.type_for` call only determines mime type based on file extension, not on content anyway. So I would suggest we just say that's good enough: if you want to use video's in your wiki, you should use the appropriate file extensions. That allows for the simplest implementation. Perhaps we can still have an optional `mime` argument for the Macro in case the user wants to override the extension-detected type.

atrost: Thanks, I will plan to incorporate the changes. Moreover, after reading the html5 spec and also based on the information from moziila here https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Video_and_audio_content, we don't really need to worry about mime types unless we want to implement some sort of optimization logic in our gollum macro as per the html spec (refer to the <source> element and its 'type' attribute). Here is the quote from the spec:  

"The type attribute of the source element allows the user agent to avoid downloading resources that use formats it cannot render."
> 

This idea elaborated further in the mozill'a documentation I linked to above. So, the usage of 'type/mime' in this context has only meaning when using multiple <source> elements giving the browser a chance to skip sources with mime types it doesn't support. According to the spec, if you don't specify a mime type for a source, the browser will need to download it first to determine the mime. Then skip to the next source, if needed to, and so forth.  So the 'type' attribute is just a hint to the browser and thus optional.  But for performance reasons, it should be used when linking to a video with multiple sources.  For wiki users it also would be simpler to just specify the file name without being bothered with mime types.

dometto: Ah, that's useful info! Looking forward to the updates 👍

atrost: Well it is now simplified to a bare minimum for what is needed to link to mp4 files in the wiki. We should recommend mp4 as this is the widely used format and all browsers support it.

gemspec.rb
```diff
@@ -8,7 +8,7 @@ def specification(version, default_adapter, platform = nil)
     s.name              = 'gollum-lib'
     s.version           = version
     s.platform          = platform if platform
-    s.date              = '2017-04-04'
+    s.date              = '2017-04-08'
```
dometto: Cleaner to remove this, but not a big problem.

lib/gollum-lib/macro/navigation.rb
```diff
@@ -6,7 +6,7 @@ def render(title = "Navigate in the TOC", toc_root_path = ::File.dirname(@page.p
         if @wiki.pages.size > 0
           list_items = @wiki.pages.map do |page|
             if page.url_path.start_with?(toc_root_path)
-              path_display = full_path ? page.url_path_display  : page.url_path_display.sub(toc_root_path,"").sub(/^\//,'')
+              path_display = full_path ? page.url_path_display  : page.url_path_display.sub(toc_root_path.gsub("-", " "), "").sub(/^\//,'')
```
dometto: Ah, this is an artifact of changing the base form `master` to `5.x`. Could you remove this change?

lib/gollum-lib/macro/video.rb
```diff
@@ -0,0 +1,9 @@
+module Gollum
+  class Macro
+    class Video < Gollum::Macro
+      def render (fname)
+        "<video width=\"100%\" height=\"100%\" src=\"#{fname}\" controls> HTML5 video is not supported on this Browser.</video>"
```
dometto: Sweet, that's a really simple implementation. I guess almost the exact same code could be used for the audio Macro.

lib/gollum-lib/version.rb
```diff
@@ -1,5 +1,5 @@
 module Gollum
   module Lib
-    VERSION = '4.2.4'
+    VERSION = '4.2.5'
```
dometto: Another problem with rebasing to the `5.x` branch. Version [should be](https://github.com/gollum/gollum-lib/blob/gollum-lib-5.x/lib/gollum-lib/version.rb#L3) `'5.0.a.3'`.

dometto:
Great! This is almost ready to merge, if you could please:
* implement the remaining cleanups indicated in the line comments
* add this change (and your name/username, if you want to) to the `HISTORY.md` file and the [5.0 Release notes](https://github.com/gollum/gollum/wiki/5.0-release-notes). :)
* squash the commits into one (if that's difficult because of the rebase let me know, I can try to squash it through the GitHub interface)

To get the Travis build to pass, I think we need to close and reopen this PR because of the rebase. Will do that in a second.

lib/gollum-lib/macro/video.rb
atrost: Yes, glad we come up with a simplified solution.  Yes we could do the same thing for the audio should be there any requests for it in the future.

lib/gollum-lib/version.rb
atrost: Done.

atrost:
I just pushed the requested cleanups to the branch. Yes please if you could do the rebase that will be helpful. Do you want me to update history.md and RN after the rebase/merge?

atrost:
Sorry just figured our there where debug artifacts that had to be removed. It's now pushed to the branch.

atrost:
@dometto. Checking back in as this pull request thread has been silent for some time now.  I have updated the 5.0 Release notes.. I was wandering if this could also be merged into the v4 branch, if so, that will be very helpful.  

Regarding `history.md` in the gollum-lib project. Currently for any 5.0 related features it is referring to 5.0 release notes (https://github.com/gollum/gollum/wiki/5.0-release-notes) in the gollum project   where I already added  information about the video macro.

Are we good to go with the merge or is there anything else you need me to change or tweak?  Thanks.

dometto:
Sorry for the wait, had some busy times. I'll have a look tonight and hopefully merge. I'll see how difficult it is to do a backport!

atrost:
Thanks!  It will be nice to have this macro available in both v4 and v5. In the meantime I have managed to assemble a solution where I can use this feature from a forked version of gollum v4 and gollum-lib v4 but I suspect it would be harder to maintain this configuration going forward.

dometto:
Unfortunately we still have one failing test on `jruby`, but it looks like a real one:
```
Failure:

  </<video (.*) (.*) src="\/Uploads\/foo.mp4" controls=""> (.*)<\/video>/> expected to be =~

  <"<p><video controls=\"\" height=\"100%\" src=\"/Uploads/foo.mp4\" width=\"100%\"> HTML5 video is not supported on this Browser.</video></p>\n">.

test_Video_macro_given_a_name_of_a_file_displays_an_html5_video_tag_(Macros)

/home/travis/build/gollum/gollum-lib/test/test_macros.rb:134:in `block in test_Video_macro_given_a_name_of_a_file_displays_an_html5_video_tag_'

     131:   test "Video macro given a name of a file displays an html5 video tag " do

     132:     file = "/Uploads/foo.mp4"

     133:     @wiki.write_page("VideoTagTest", :markdown, "<<Video(#{file})>>", commit_details)

  => 134:     assert_match /<video (.*) (.*) src="#{file}" controls=""> (.*)<\/video>/, @wiki.pages[0].formatted_data

     135:   end 

     136: end
```
If I'm not mistaken the issue is that the `controls` bit in the actual output is *before* the `src` bit, and it's expecting the reverse. Any idea why this might be failing on `jruby`?

atrost:
From the browser's perspective there is no predefined order for the attributes in html so any sequence is allowed, so the generated html should work in either case (the src bit before or after the controls bit etc), but for the rspec the order is obviously matters. If that is the issue for jruby, why not simply use a more generic matcher? So instead of : ` assert_match /<video (.*) (.*) src="#{file}" controls=""> (.*)<\/video>/`   do this: `assert_match /<video (.*) (.*) src="#{file}" (.*)> (.*)<\/video>/`.
This passes the ruby rspec. Unfortunetely, I cannot currently install Java or jruby env on my development system so cannot verify this. Could you please try this in your env to see if this works?  I can push the revised rspec with the suggested modification, if needed to. Thanks.

dometto:
Good idea :+1:

atrost:
I pushed some change if you can please try it out.

Are there any specific jruby requirements for Gollum Wiki, like jruby version, compatible java app servers? Thanks

dometto:
Still getting the same failure, as you can see [on Travis](https://travis-ci.org/gollum/gollum-lib/builds/238819324?utm_source=github_status&utm_medium=notification). The issue is still the order of the attributes. I'm in favor of just removing the `controls` bit from the expected match.

dometto:
Sorry for the long wait and thanks for the Macro, @atrost!

PR #279, opened 2017-06-12
Update Nokogiri gem to a non-vulnerable version. Fixes #278

### LINKED ISSUE #278: gollum-lib linked to vulnerable nokogiri version
Hi,

`nokogiri` dependencies are tightened to `~> 1.6.4`, however 1.6.x includes vulnerable bundled libs:

```
Name: nokogiri
Version: 1.6.8.1
Advisory: CVE-2016-4658
Criticality: Unknown
URL: https://github.com/sparklemotion/nokogiri/issues/1615
Title: Nokogiri gem contains several vulnerabilities in libxml2 and libxslt
Solution: upgrade to >= 1.7.1

Vulnerabilities found!
```

d2bit:
Fixes #278 
Updated Nokogiri to > 1.7.1

Test suite run output:
```
Started
................................................................................................................................................................
..................................................................................................................

Finished in 14.667878 seconds.
----------------------------------------------------------------------------------------------------------------------------------------------------------------
274 tests, 635 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
----------------------------------------------------------------------------------------------------------------------------------------------------------------
18.68 tests/s, 43.29 assertions/s
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,8 @@ rvm:
   - 2.3.0
   - 2.4.0
   - jruby-19mode
+jdk:
+  - oraclejdk8
 before_install:
  - sudo apt-get update
  - sudo apt-get install libicu-dev
--- a/gemspec.rb
+++ b/gemspec.rb
@@ -26,7 +26,7 @@ def specification(version, default_adapter, platform = nil)
 
     s.add_dependency *default_adapter
     s.add_dependency 'rouge', '~> 2.0'
-    s.add_dependency 'nokogiri', '~> 1.6.4'
+    s.add_dependency 'nokogiri', '~> 1.7', '>= 1.7.1'
     s.add_dependency 'stringex', '~> 2.6'
     s.add_dependency 'sanitize', '~> 2.1'
     s.add_dependency 'github-markup', '~> 1.6'
--- a/gemspec.rb
+++ b/gemspec.rb
@@ -26,7 +26,11 @@ def specification(version, default_adapter, platform = nil)
 
     s.add_dependency *default_adapter
     s.add_dependency 'rouge', '~> 2.0'
-    s.add_dependency 'nokogiri', '>= 1.6.1', '< 2.0'
+    if RUBY_VERSION < '2.1'
+      s.add_dependency 'nokogiri', '~> 1.6.1'
+    else
+      s.add_dependency 'nokogiri', '>= 1.6.1', '< 2.0'
+    end
     s.add_dependency 'stringex', '~> 2.6'
     s.add_dependency 'sanitize', '~> 2.1'
     s.add_dependency 'github-markup', '~> 1.6'
--- a/gemspec.rb
+++ b/gemspec.rb
@@ -26,7 +26,7 @@ def specification(version, default_adapter, platform = nil)
 
     s.add_dependency *default_adapter
     s.add_dependency 'rouge', '~> 2.0'
-    s.add_dependency 'nokogiri', '~> 1.7', '>= 1.7.1'
+    s.add_dependency 'nokogiri', '>= 1.6.1', '< 2.0'
     s.add_dependency 'stringex', '~> 2.6'
     s.add_dependency 'sanitize', '~> 2.1'
     s.add_dependency 'github-markup', '~> 1.6'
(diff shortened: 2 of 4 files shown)


josacar:
@bartkamphorst Is there any show stopper to merge this?

bartkamphorst:
Yes, there is a test failing on Travis.

josacar:
@bartkamphorst This is nothing about nokoguiri, it's aboute `rouge` syntax highlighter:

```
Installing rouge 2.0.7 (was 2.1.1)
```

and then

```
Installing rouge 2.0.7 (was 2.1.1)
Using gollum-lib 4.2.5 from source at `.`
Bundle updated!
~/code/gollum-lib (master)$ be ruby test/test_markup.rb
Loaded suite test/test_markup
Started
.................................................................................................

Finished in 4.821316 seconds.
---------------------------------------------------------------------------------------------------------------------------------------------------------
97 tests, 168 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
---------------------------------------------------------------------------------------------------------------------------------------------------------
20.12 tests/s, 34.85 assertions/s
```
 
Should we update specs or should we pin `rouge` gem to `2.0.7` ?

Thanks in advance.

d2bit:
@josacar good catch! I was not able to reproduce the error, as my Gemfile.lock pointed to 2.0.7.
@bartkamphorst I think master builds are red now. Do you prefer to pin rouge version, or to update the specs?

bartkamphorst:
@d2bit @josacar I prefer updating the spec. 👍 Thanks for working on this!

josacar:
@bartkamphorst Everything green but MRI 2.0 and JRuby

bartkamphorst:
I hadn't realized that nokogiri 1.7 requires Ruby version >= 2.1.0. This breaks backwards compatibility. @dometto What do you suggest we do? Release a minor release or include this fix only in 5.x? Given the security risk I think having the in 4.X makes some sense.

josacar:
Maybe we can relax Nokogiri dependency to < 2.0 in 4.x branch and let everyone choose to stay updated or run with an EOL version and vulnerable. What do you think?

dometto:
@josacar good suggestion. If it means  breaking the MRI < 2.1.0 test on Travis, I say we just remove that.

josacar:
Tried loosing the dependency to < 2.0, but rubygems still requires latest compatible version although it's not compatible with the MRI version.

josacar:
@dometto @bartkamphorst I've modified the gemspec to make MRI 2.0 pass on CI, Jruby is still failing due to a dependency is missing

josacar:
New JGit requires Java 8, fixed in travis

dometto:
Sorry for the wait, I was away for a bit. Merging this now. I'll also bump the gem.

dometto:
@josacar just FYI: I overlooked this, but the `if` statement in `gemspec.rb` doesn't work as you might expect. As the code is only evaluated on build-time (i.e., when running `gem build`), the right `nokogiri` version is not required dynamically. Rather, you just end up with a different gem depending on which version of ruby you're using. This had the effect of breaking `ruby 2.0.0` support for the `gollum` gem. No big problem, as I think we should just decontinue support for that ruby, seeing as it's not compatible with a safe version of `nokogiri`. But I just wanted to give you a heads up about this annoying feature of ruby Gemfiles.

josacar:
Ok, my bad. We can still do a `nokogiri "~> 1.6" and do the if statement to make CI pass, and let users update to latest nokogiri whilst letting users use an unsafe nokogiri version. What do you think?

josacar:
In fact, if you're building with > 2.0 it should happen nothing, as it requires `>= 1.6.1`

dometto:
I actually think it is ok to make installation on `2.0` fail so nobody uses a vulnerable gem by accident. Currently gollum-lib depends on `nokogiri v1.8`, but a `gem install gollum` will still work because it will require an older `gollum-lib`. I think the thing to do is bump the `gollum` gem and make it depend strictly on the latest, non-vulnerable `gollum-lib`.

bartkamphorst:
👍

Issue #280, opened 2017-06-29
GlobalTOC() Macro doesn't support base_path


Agilus:
If you start gollum with the --base_path option, giving you a wiki that is not at the root of a URL (e.g. http://myhost.com/subpathwiki instead of http://myhost.com/), the GlobalTOC() creates a list of links that point to the wrong place.  

Instead of pointing to http://myhost.com/subpathwiki/Home, a link made by GlobalTOC() will point to http://myhost.com/Home .

The fix for this problem is relatively simple, and I have it ready to commit.

irstevenson:
Same issue (and by the looks of it, same fix need) with `Navigation()`.

dometto:
Fixed in `v5.x`

PR #281, opened 2017-06-29
Fixed issue with GlobalTOC() not working with a base_path other than / .

Agilus:
This was issue #280 on the main gollum-lib project.

### DIFF
--- a/lib/gollum-lib/macro/global_toc.rb
+++ b/lib/gollum-lib/macro/global_toc.rb
@@ -3,7 +3,9 @@ module Gollum
     class GlobalTOC < Gollum::Macro
       def render(title = "Global Table of Contents")
         if @wiki.pages.size > 0
-          result = '<ul>' + @wiki.pages.map { |p| "<li><a href=\"/#{p.url_path}\">#{p.url_path_display}</a></li>" }.join + '</ul>'
+          # Strip trailing slash.  This is important if base_path is /
+          prepath=@wiki.base_path.sub(/\/$/, '')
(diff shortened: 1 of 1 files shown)


dometto:
👍 Thanks. However, there are some test failures. Don't really seem to be related (or are they)? Would be great if you could fix that.

Agilus:
Yeah, I saw that some tests didn't pass.  I will see if I can correct the problem!

Agilus:
Changed "wiki link within org code block" test in test/test_markup.rb to include an extra span tag.  This seems like a pretty benign change.  However, it doesn't seem to be related to the change I made to the global_toc macro.

One of the tests run by Travis-CI is still failing, but it looks like a configuration problem, and not related to the code at all (but I could be wrong).

deric:
@dometto The Travis failure doesn't seem to be related to this change (the problem was probably fixed by commit 65a8762aa47ec0fe0df0f6bfd79a3481352efecb). Is this ready for merge?

dometto:
Sorry for the slow response, was away for a bit. I'll restart the travis build and see if we can merge this without breaking the build.

dometto:
My apologies. I changed the target branch to `5.x` and that caused another test failure, because in that branch we have refactored-away the `Page#url_path_display` method. The call to that method in `global_toc.rb` should just be changed to `#url_path`. Could you please update the branch in that way?

dometto:
Ping @Agilus

dometto:
Also, it would be great if you could implement the same fix for the `Navigation` macro in this PR!

Agilus:
I'll try to take a look at this in the next few days.  Management shifted direction to using a new product, so I've been checked out on Gollum, but I can try to get this finished up.

PR #283, opened 2017-08-16
resubmitting Fix issue #1161/#232 rendering of TOC on the sidebar

karlredman:
Forgive me if I've missed some steps: I haven't submitted actual pull requests for a github project before. Please advise if I should be offering more information or proof of the fix. 

This is an attempt to resubmit [a pull request originated by @bogado](https://github.com/gollum/gollum-lib/pull/232/commits/e19b4f5363d84f64beb53bbc29468d75b693ee1d).

I have personally tested this with gollum v4.1.2+ But I do not know how I would go about writing a unit test for this.

### DIFF
--- a/lib/gollum-lib/page.rb
+++ b/lib/gollum-lib/page.rb
@@ -495,7 +495,9 @@ module Gollum
           end
           
           if subpageblob
-            instance_variable_set("@#{subpagename}", subpageblob.page(@wiki, @version) )
+            subpage =  subpageblob.page(@wiki, @version)
+            subpage.parent_page = self
+            instance_variable_set("@#{subpagename}", subpage)
             break
           end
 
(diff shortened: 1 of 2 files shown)


dometto:
Great, thanks! I think to test for this all we need to do is add this assertion to `test_page.rb:`

```ruby
assert page.sidebar.parent_page = page
```
...in the test `test "should inherit header/footer/sidebar pages from parent directories"`.

karlredman:
Ah, cool! Sorry for the delay. I'm looking at this today to make sure I understand the unit tests. Thanks! I'll submit this sometime today.

karlredman:
ok, I finally got this up and running i think. Sorry for the delay -I had a complete system re-install in the mean time and learned a lot about ruby. Let me know if This code/test is all good. Much thanks for your patience and help.

PR #287, opened 2017-10-05
Verbose search. Resolves #166

### LINKED ISSUE #166: Better Search
At the moment search returns page names and number of matches, this makes finding what you're looking for still rather painful. It would be much better if it could return the actual matched content.

I understand this is probably a challenge because you don't want to be returning the raw source to the end user, and if you cut out a line it may not parse properly if you want to attempt to convert it to html. It's a challenge but imo it's one that'll need to be overcome.

C-Abner:
modify something for "verbose search" feature.

just like 
image

also modify some files on gollum,  gollum-lib and gollum-grit_adapter.

### DIFF
--- a/lib/gollum-lib/wiki.rb
+++ b/lib/gollum-lib/wiki.rb
@@ -641,14 +641,24 @@ module Gollum
     # Returns an Array with Objects of page name and count of matches
     def search(query)
       options = {:path => page_file_dir, :ref => ref}
+
       results = {}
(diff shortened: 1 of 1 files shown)


dometto:
See comments in https://github.com/gollum/grit_adapter/pull/18/files

dometto:
Ping @C-Abner

dometto:
Closing for lack of response, in favor of #343

PR #292, opened 2018-01-15
Some fixes when moving a page using a multi committer

fjsanpedro:
I'm trying to simulate a move (rename) operation that also can update the content of a page using the same commit, and, therefore, the same commiter.

I have realized that there are a couple of things that should be done in order to implement this. If we want to move a page we have to check first if the page has been renamed or modified in the committer tree. And, according to that information pick the right information to perform the `rename_page` operation.

### DIFF
--- a/lib/gollum-lib/wiki.rb
+++ b/lib/gollum-lib/wiki.rb
@@ -396,8 +396,17 @@ module Gollum
       multi_commit = !!commit[:committer]
       committer    = multi_commit ? commit[:committer] : Committer.new(self, commit)
 
+      # This piece only works for multi_commit
+      # If we are in a commit batch and one of the previous operations
+      # has updated the page, any information we ask to the page can be outdated.
+      # Therefore, we should ask first to the current committer tree to see if
+      # there is any updated change.
+      raw_data = raw_data_in_commiter(committer, source_dir, page.filename) ||
+                 raw_data_in_commiter(committer, source_dir, "#{target_name}.#{Page.format_to_ext(page.format)}") ||
+                 page.raw_data
+
       committer.delete(page.path)
-      committer.add_to_index(target_dir, target_name, page.format, page.raw_data)
+      committer.add_to_index(target_dir, target_name, page.format, raw_data)
 
       committer.after_commit do |index, _sha|
         @access.refresh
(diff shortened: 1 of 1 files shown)


lib/gollum-lib/wiki.rb
```diff
@@ -431,7 +440,7 @@ def rename_page(page, rename, commit = {})
     # Returns the String SHA1 of the newly written version, or the
     # Gollum::Committer instance if this is part of a batch update.
     def update_page(page, name, format, data, commit = {})
-      name     ||= page.name
+      name     = name.present? ? ::File.basename(name) : page.name
```
fjsanpedro: If the method update_page can only rename files, let's stay strict to that and disallow moving the file in the current directory tree. If we do this, we don't have to deal with other problems in the rename_page operation.

dometto:
@fjsanpedro sorry for the late reply. Can you explain why this change is necessary? It's difficult to see what effect this is exactly going to have on gollum's behavior.

fjsanpedro:
@dometto let me recap because it's has been some time.

Imagine that we have a wiki page that we have updated the wiki content and also moved the page to a different location (even with a different name).

I don't remember now why, but the `update` operation wasn't enough to do this. So, to perform this with only one operation (and avoid some flood in the commit history) we had to use a a `multi_commit`.

The first operation in the `multi_commit` would be the `update` and after the `rename`. When we reached the `rename` operation the content retrieved was from `page.raw_data` but that is wrong because it is outdated and the updated information is in the `multi_commit`. 

This is the our `update_page` version:
```ruby
def gollum_update_page(page_path, title, format, content, commit_details)
  assert_type!(format, Symbol)
  assert_type!(commit_details, CommitDetails)

  page = gollum_page_by_path(page_path)
  committer = Gollum::Committer.new(page.wiki, commit_details.to_h)

  # Instead of performing two renames if the title has changed,
  # the update_page will only update the format and content and
  # the rename_page will do anything related to moving/renaming
  gollum_wiki.update_page(page, page.name, format, content, committer: committer)
  gollum_wiki.rename_page(page, title, committer: committer)
  committer.commit
  nil
end
```

If you still find it difficult to see the whole picture or how it can interfere with the other lib code, I can dig deeper about our reasoning at that time, but it will take me some time.

Thanks for looking into this 😉

dometto:
@fjsanpedro I'm happy to merge something like this in if it helps on your end, but:

* The tests are failing (and there seem to be no tests for the new behavior)
* We're only actively developing against the `gollum-lib-5.x` branch

I'll close this for now but feel free to (re)open a new PR!

PR #296, opened 2018-03-23
Upgrade Sanitize to 4.6.4 to fix CVE-2018-3740

rgrove:
Sanitize <= 4.6.2 has an HTML injection vulnerability due to unsafe behavior in libxml2. This upgrades Sanitize to 4.6.4, which is not vulnerable.

See https://github.com/rgrove/sanitize/issues/176 for more details. Please let me know if you have any questions.

Note: Two tests failed on my machine before this patch. The same two tests fail in the same way after this patch, so I'm calling that a success.

### DIFF
--- a/gemspec.rb
+++ b/gemspec.rb
@@ -32,7 +32,7 @@ def specification(version, default_adapter, platform = nil)
       s.add_dependency 'nokogiri', '>= 1.6.1', '< 2.0'
     end
     s.add_dependency 'stringex', '~> 2.6'
-    s.add_dependency 'sanitize', '~> 2.1'
+    s.add_dependency 'sanitize', '~> 4.6.4'
     s.add_dependency 'github-markup', '~> 1.6'
     s.add_dependency 'gemojione', '~> 3.2'
 
--- a/lib/gollum-lib/filter/sanitize.rb
+++ b/lib/gollum-lib/filter/sanitize.rb
@@ -8,7 +8,7 @@ class Gollum::Filter::Sanitize < Gollum::Filter
   def process(data)
     if @markup.sanitize
       doc = Nokogiri::HTML::DocumentFragment.parse(data)
-      doc = @markup.sanitize.clean_node!(doc)
+      doc = @markup.sanitize.node!(doc)
 
       doc.to_xml(@markup.to_xml_opts)
     else
--- a/lib/gollum-lib/page.rb
+++ b/lib/gollum-lib/page.rb
@@ -121,7 +121,7 @@ module Gollum
     #
     # Returns the fully sanitized String title.
     def title
-      Sanitize.clean(name).strip
+      Sanitize.fragment(name).strip
     end
 
     # Public: Determines if this is a sub-page
--- a/lib/gollum-lib/sanitization.rb
+++ b/lib/gollum-lib/sanitization.rb
@@ -150,7 +150,7 @@ module Gollum
       end
     end
 
-    # Builds a Hash of options suitable for Sanitize.clean.
+    # Builds a Hash of options suitable for Sanitize.fragment.
     #
     # Returns a Hash.
     def to_hash
--- a/test/test_markup.rb
+++ b/test/test_markup.rb
@@ -16,7 +16,7 @@ context "Markup" do
     assert @wiki.pages[0].formatted_data
   end
 
-  # This test is to assume that Sanitize.clean doesn't raise Encoding::CompatibilityError on ruby 1.9
+  # This test is to assume that Sanitize.fragment doesn't raise Encoding::CompatibilityError on ruby 1.9
   test "formats non ASCII-7 character page from Wiki#pages" do
     wiki = Gollum::Wiki.new(testpath("examples/yubiwa.git"))
     assert_nothing_raised(defined?(Encoding) && Encoding::CompatibilityError) do


rgrove:
Actually, I see now in Travis that gollum supports JRuby. Sanitize >= 3.0.0 isn't currently compatible with JRuby since Sanitize depends on the native nokogumbo library for HTML parsing. Not sure how you want to proceed here.

captn3m0:
Relevant issue on nokogumbo: https://github.com/rubys/nokogumbo/issues/24

I don't see that getting resolved anytime soon. Is dropping jruby support an option?

tnir:
@rgrove While I do not think so, what do you mean?

> Note: Two tests failed on my machine before this patch. The same two tests fail in the same way after this patch, so I'm calling that a success.

The latest build on master is https://travis-ci.org/gollum/gollum-lib/builds/290365658, which has some network? error only in Ruby 2.4
and the build on your source branch is https://travis-ci.org/gollum/gollum-lib/builds/357230753, which has errors in all jobs.

captn3m0:
What's the next step here?

tnir:
@captn3m0 As a user, we can use gitlab-gollum-lib instead. It is provided by GitLab, Inc. Replace `gollum-lib` with `gitlab-gollum-lib` in your Gemfile.

- Rubygem: https://rubygems.org/gems/gitlab-gollum-lib
- Project repo (forked): https://gitlab.com/gitlab-org/gollum-lib

captn3m0:
Thanks for the tip. Going to try if this works with our setup.

dometto:
Sorry for the (too) slow response here. The two test failures are not a show stopper for merging this, but the breaking of JRuby support is. Any recommendations on how to get Sanitize to play with JRuby?

cc @bartkamphorst

rgrove:
@dometto I'm afraid there's not a path forward with Sanitize 4.x if you need JRuby support. You may want to consider backporting [this fix](https://github.com/rgrove/sanitize/commit/01629a162e448a83d901456d0ba8b65f3b03d46e) to the [2.x line](https://github.com/rgrove/sanitize/tree/v2.1.0), which is the last version of Sanitize that's compatible with JRuby.

dometto:
Thanks @rgrove. Would you also accept a patch to sanitize `4.x` that restores `jruby` compatibility by only conditionally requiring the c-extension that breaks it? I haven't looked yet if that functionality can be implemented (more slowly, I suppose) without the c-extensions, but we've had some success in providing `jruby` support in this way, for instance in `github-markup`.

rgrove:
@dometto I would happily accept a patch to 4.x that adds JRuby compatibility, but the C library in question is the Gumbo HTML parsing library, which is crucial to Sanitize's core functionality. I don't think removing it will be an option, but if you have any ideas I'm all ears.

dometto:
Thanks @rgrove.

I have no suggestions other than falling back on `nokogiri` on `JRuby`. Was you decision to move away from that based on performance or security concerns? I know there are issues about XHTML compliance and consistency with the ruby/c version of `nokogori`. The `JRuby` version of `nokogori` is, however, a different implementation altogether, based on [nekoHTML](http://nekohtml.sourceforge.net/) and [Apache Xerces](https://xerces.apache.org/#Apache+Xerces+Project+-+Current+Projects), so it *may* be better. If `Sanitize` allows easy switching of parsing engines, I'd be happy to try and get it working with `JRuby`-based `nokogiri`. (Although it's not ideal -- probably some specs will give different results with different parsing engines)

In the mean time, we should probably backport the patch to `2.x`. Would you accept a PR for that and update the `2.x` gem?

dometto:
@bartkamphorst another long term solution would be two switch to [this](https://github.com/flavorjones/loofah) or [this](https://github.com/flavorjones/loofah)

rgrove:
@dometto The only reason I switched from Nokogiri's parser to Gumbo was that I needed a parser that followed the HTML parsing spec and supported modern HTML so that Sanitize's parsing behavior aligned more closely with browser parsing behavior.

> If Sanitize allows easy switching of parsing engines, I'd be happy to try and get it working with JRuby-based nokogiri. (Although it's not ideal -- probably some specs will give different results with different parsing engines)

It shouldn't be too hard to get Sanitize working with a non-Gumbo parser since it still uses Nokogiri to manipulate parsed documents, though the devil will be in the details. As a start, you'd need to:

1. Conditionally [require `nokogiri` instead of `nokogumbo`](https://github.com/rgrove/sanitize/blob/cee5bfa9dfa00a9f4fd11f598039cd47b997d585/lib/sanitize.rb#L3).

2. Conditionally [call `Nokogiri::HTML.parse` instead of `Nokogiri::HTML5.parse`](https://github.com/rgrove/sanitize/blob/cee5bfa9dfa00a9f4fd11f598039cd47b997d585/lib/sanitize.rb#L111-L124).

You'd then need to look for possible edge cases that rely on assumed HTML5 parsing behavior during sanitization, though I don't think there are likely to be many. There will definitely be many test failures though, since the parser output will be different.

If someone is willing to put some work into this, I'd be happy to accept pull requests. The only reason I haven't done it myself is a lack of time and no personal use case for JRuby support.

> In the mean time, we should probably backport the patch to 2.x. Would you accept a PR for that and update the 2.x gem?

Happily!

dometto:
@rgrove after looking into this a bit I think accounting for the differences in output between `nokogiri` and `nokogumbo` would be too hacky. I'm getting about 41 test failures when I replace `Nokogiri::HTML5` with `Nokogiri::HTML` on `jruby`. The right way to fix this would be to make `nokogumbo` itself `jruby`-compatibile, of which there is talk [here](https://github.com/rubys/nokogumbo/issues/24).

For `gollum v5` we can switch to using `loofah`, at least as long as `nokogumbo` isn't compatible yet. That means we have to concentrate on backporting the fix now, so it gets rolled out automatically with a `gem update gollum` for those running `4.x`.

Unfortunately it seems as if the relevant file in `sanitize` (clean_element.rb) has undergone a substantial rewrite since `v2.1.0`. I'm not sure what the best way to proceed would be: I've perhaps naively tried to just checkout the current version of `clean_element.rb`, but that triggered a bunch of errors. Alternatively, is there a more abstract writeup of what needs to happen in order to address the vulnerability that I could try and implement on `v2.1.0`?

rgrove:
@dometto It's definitely not feasible to backport entire files from the 4.x line to the 2.x line. This fix will need to be reimplemented on top of the 2.x line, but it's not complicated.

[This libxml2 bug](https://bugzilla.gnome.org/show_bug.cgi?id=769760) describes the root of the problem. Essentially, libxml2 allows unencoded space and double-quote characters in the values of certain attributes whose values should be URI-encoded (`name` on `<a>` elements and `action`, `href`, or `src` on all elements). This can allow an attacker to inject HTML by prematurely closing the attribute with an unencoded space or double-quote. To fix this, we must URI-encode space and double-quote characters in these attributes.

This is the commit that fixes this issue in the Sanitize 4.x line: https://github.com/rgrove/sanitize/commit/01629a162e448a83d901456d0ba8b65f3b03d46e

The most relevant portion is here: https://github.com/rgrove/sanitize/blob/cee5bfa9dfa00a9f4fd11f598039cd47b997d585/lib/sanitize/transformers/clean_element.rb#L160-L175

After all other filtering has been performed on an attribute, that code checks whether the attribute is one of the ones that needs additional escaping, and then performs that escaping if it is.

The tests added in https://github.com/rgrove/sanitize/commit/01629a162e448a83d901456d0ba8b65f3b03d46e should also be backported to the 2.x line to confirm the fix.

PR #299, opened 2018-09-24
Switch to GHF markdown by default

dometto:
Switch to GFM by default. See https://github.com/gollum/gollum/issues/1308#issuecomment-423969590


dometto:
@bartkamphorst: as you can see the switch is pretty painless. Your call on whether we want to make this default or not!

bartkamphorst:
Seems to me that staying somewhat close to Github markdown is still favorable, so go for it!

PR #302, opened 2018-10-03
Update kramdown and disable kramdown math handling

dometto:
`kramdown`'s own attempt to handle math was inserting `<script>` tags into the rendered data which was then being sanitized, thus breaking math support altogether. Also see https://github.com/gollum/gollum/pull/1325


bartkamphorst:
Nice catch. :+1:

PR #303, opened 2018-10-13
Make path lookup to be case sensitive

0sc:
Fixes https://github.com/gollum/gollum/issues/824


dometto:
Looks good to me! Thanks @0sc

PR #304, opened 2018-10-14
Prepend base_path to links in Navigation()

dometto:
See #280


bartkamphorst:
Nice. 👍

dometto:
Also caught two bugs:
* the Macro would previously match any part that *started with* the designated root path, so if you wanted to list all pages under `Gondor`, all pages under `Gondorlongerpath` would also match.
* the Macro was substituting out all the dots in the displayed filenames, because it was removing all instances of `toc_root_path` in the filename -- so when `toc_root_path` is `'.'` it was subbing out the dots.

Also added tests for this macro.

PR #305, opened 2018-10-14
Do not display metadata if only keys are title or countheads

dometto:
Related to https://github.com/gollum/gollum/pull/1330


bartkamphorst:
@dometto Good to go!

PR #306, opened 2018-10-19
Implement global metadata

dometto:
In `gollum` this will allow defining `wiki_options[:metadata]` in a `config.rb` in order to set global metadata.

### DIFF
--- a/lib/gollum-lib/page.rb
+++ b/lib/gollum-lib/page.rb
@@ -225,7 +225,7 @@ module Gollum
     def metadata
       unless @metadata
         formatted_data if markup.metadata == nil
-        @metadata = (markup.metadata || {}).merge(@wiki.metadata)
+        @metadata = @wiki.metadata.merge(markup.metadata || {})
       else
         @metadata
       end
--- a/lib/gollum-lib/page.rb
+++ b/lib/gollum-lib/page.rb
@@ -223,8 +223,13 @@ module Gollum
     #
     # Returns Hash of metadata.
     def metadata
-      formatted_data if markup.metadata == nil
-      markup.metadata || {}
+      unless @metadata
+        formatted_data if markup.metadata == nil
+        @metadata = markup.metadata || {}
+        @metadata.merge(@wiki.metadata)
+      else
+        @metadata
+      end
     end
 
     # Public: The format of the page.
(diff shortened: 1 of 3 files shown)


bartkamphorst:
Nice. 👍

lib/gollum-lib/page.rb
```diff
@@ -223,8 +223,12 @@ def toc_data
     #
     # Returns Hash of metadata.
     def metadata
-      formatted_data if markup.metadata == nil
-      markup.metadata || {}
+      unless @metadata
+        formatted_data if markup.metadata == nil
```
bartkamphorst: Why is this statement within the `unless` block?

dometto: I was thinking it needs to be executed only once, when `@metadata` is not yet initialized

dometto: The call is just to ensure that the page's metadata has actually been parsed

bartkamphorst: Remind me what `formatted_data` does?

dometto: It returns the page's contents after being parsed through the filter chain (which is also where metadata, table of contents, etc. are being generated).

lib/gollum-lib/page.rb
```diff
@@ -225,7 +225,7 @@ def toc_data
     def metadata
       unless @metadata
         formatted_data if markup.metadata == nil
-        @metadata = (markup.metadata || {}).merge(@wiki.metadata)
+        @metadata = @wiki.metadata.merge(markup.metadata || {})
```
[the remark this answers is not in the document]
bartkamphorst: > do you agree that's the most intuitive behavior?

Yes! Nice catch.

lib/gollum-lib/page.rb
dometto: It's to initialize `markup.metadata`. Prior to calling `markup.render`, the page's contents won't yet have been parsed, and so `markup.metadata` returns `nil`. So in constructing the page's metadata, we have to ensure that the underlying markup object has executed the filter chain.

(`formatted_data` itself stores the results in `@formatted_data`, so we don't end up rendering the contents multiple times)

bartkamphorst: Got it. Maybe add a comment there to explain this?

dometto: Yes, it is kind of counterintuitive 👍

PR #307, opened 2018-10-20
Fix encodings for filenames

dometto:
Turns out `ERB::url_encode` is [forcing](https://apidock.com/ruby/ERB/Util/url_encode#) ASCII-8bit, which is causing https://github.com/gollum/gollum/issues/1339

### DIFF
--- a/lib/gollum-lib/file.rb
+++ b/lib/gollum-lib/file.rb
@@ -39,7 +39,7 @@ module Gollum
     #
     # Returns the String url_path
     def escaped_url_path
-      ERB::Util.url_encode(self.url_path).gsub('%2F', '/')
+      ERB::Util.url_encode(self.url_path).gsub('%2F', '/').force_encoding('utf-8')
     end
 
     # Public: The on-disk filename of the file.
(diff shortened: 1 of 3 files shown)


dometto:
> Yes, I had noticed that too, but I figured that was part of what the encoding was supposed to do.

Looking at the source for `url_encode` I think it's more a necessary condition for the encoding operation.

dometto:
I think this may be happening because `@base_url` in gollum is set through a call to sinatra's `url` helper, which probably gets the base url from rack's path info -- which probably is ascii-encoded. Adding a `force_encode('utf-8')` to [this line](https://github.com/gollum/gollum/blob/5.x/lib/gollum/app.rb#L89), together with this PR, resolves the problem for me. Would that be an acceptable fix?

In general, it seems forcing input to utf-8 at entry point in the app, and then dealing only with utf-8 internally, seems to be the recommended practice ([example](http://graysoftinc.com/character-encodings/what-ruby-19-gives-us)). What I'm unclear on is whether it's best to just call `force_encoding`, or whether we should somehow be trying to transcode the strings (using `encode`).

PR #310, opened 2018-11-15
Kramdown header location marking

dometto:
To implement a clean way of editing per section (see https://github.com/gollum/gollum/issues/1000), we need the rendering gem to support sourcemaps (i.e., outputting the position in the markup source of a rendered element in the html), minimally for header elements.

`commonmarker` has [in-built sourcemapping](https://github.com/gjtorikian/commonmarker#passing-options), but for gollum I think we want to stick with `kramdown` as a default for now (`commonmarker` uses c-extensions). So:

This experimental PR overrides the default `kramdown` converter to output the position (linenumber) of a header element in the `data-sourcepos` attribute of the `hX` element. Example output (taken from the frontend):
```
 <h1 data-sourcepos="1"><a class="anchor" id="bilbo-baggins" href="#bilbo-baggins"><i class="fa fa-link"></i></a>Bilbo Baggins</h1>
```

It is slightly hackish, in so far as updates to `kramdown` might break this -- but the changes required are fairly minimal, as you can see. Maybe it wouldn't cause too much trouble, assuming that `kramdown` doesn't end up breaking SemVer. But do let me know if you think it's just too much of a hack.

In the frontend, we could simply pass the value of the `data-sourcepos` attribute along to the edit view, if it exits, and make the editor jump to that line in the source.

Complication: of course, the `render` filter is last in the filter chains. So if previous filters extract parts of the markup source, the line numbers generated by kramdown may be incorrect. Could we make it a policy for filters to always substitute back the same amount of lines as they extract?

`commonmarker` also stores the source position in the `data-sourcepos` attribute, so we could enable this when `commonmarker` is used and get the same effect in the frontend. Although the output looks slightly different:
```ruby
2.4.1 :001 > require 'commonmarker'
 => true 
2.4.1 :002 > CommonMarker.render_html("\"'Shelob' is my name.\"", [:HARDBREAKS, :SOURCEPOS])
 => "<p data-sourcepos=\"1:1-1:22\">&quot;'Shelob' is my name.&quot;</p>\n" 
```
(We only need the starting line number)

### DIFF
--- a/lib/gollum-lib/markups.rb
+++ b/lib/gollum-lib/markups.rb
@@ -41,6 +41,16 @@ end
 
 include Gollum::MarkupRegisterUtils
 
+if gem_exists?('kramdown') then
+  require 'kramdown'
+  class Kramdown::Converter::HtmlSourcemap < Kramdown::Converter::Html
+    def convert_header(el, indent)
+      el.attr[:'data-sourcepos'] = el.options[:location]
+      super(el, indent)
+    end
+  end
+end
+
 module GitHub
   module Markup
     class Markdown < Implementation
@@ -56,7 +66,7 @@ end
 module Gollum
   class Markup
     GitHub::Markup::Markdown::MARKDOWN_GEMS['kramdown'] = proc { |content|
-        Kramdown::Document.new(content, :input => "GFM", :auto_ids => false, :math_engine => nil, :smart_quotes => ["'", "'", '"', '"'].map{|char| char.codepoints.first}).to_html
+      Kramdown::Document.new(content, :input => "GFM", :auto_ids => false, :math_engine => nil, :smart_quotes => ["'", "'", '"', '"'].map{|char| char.codepoints.first}).to_html_sourcemap
     }
     GitHub::Markup::Markdown::MARKDOWN_GEMS['pandoc-ruby'] = proc { |content|
         PandocRuby.convert(content, :s, :from => :markdown, :to => :html, :filter => 'pandoc-citeproc')
--- a/lib/gollum-lib/sanitization.rb
+++ b/lib/gollum-lib/sanitization.rb
@@ -223,6 +223,12 @@ module Gollum
     # Default whitelisted attributes.
     ATTRIBUTES = ({
         'a'   => ['href'],
+        'h1'  => ['data-sourcepos'],
+        'h2'  => ['data-sourcepos'],
+        'h3'  => ['data-sourcepos'],
+        'h4'  => ['data-sourcepos'],
+        'h5'  => ['data-sourcepos'],
+        'h6'  => ['data-sourcepos'],
         'img' => ['src'],
         :all  => ['abbr', 'accept', 'accept-charset',
                   'accesskey', 'action', 'align', 'alt', 'axis',


bartkamphorst:
I'm really not familiar with data-sourcepos, so I'm having some difficulty understanding the proposal. In general, though, wouldn't kramdown want to support this?

dometto:
`data-sourcepos` is just an arbitrary HTML attribute to store the information in. Kramdown already stores the `location` (=starting line number) for most of its elements, so it's just a matter of making sure it ends up in the rendered HTML.

I also wondered whether this might be accepted as a feature in kramdown (although then it would have to be done for all elements, not just

dometto:
What do you think of the prospects of keeping the original linenumbering intact in the rest of the filters, though?

bartkamphorst:
> Could we make it a policy for filters to always substitute back the same amount of lines as they extract?

I don't think this will play nicely with Macros.

dometto:
Why the Macros specifically?

I was thinking that what we'd need to ensure is just that the placeholders that get put in in the `extract` fase span the same number of lines as the snippet that was extracted. What happens after the render fase, e.g. the execution of Macros, is of no concern for making sure the linenumbering matches up with what's shown to the user in the edit view

bartkamphorst:
>  I was thinking that what we'd need to ensure is just that the placeholders that get put in in the extract phase span the same number of lines as the snippet that was extracted. What happens after the render phase, e.g. the execution of Macros, is of no concern for making sure the linenumbering matches up with what's shown to the user in the edit view

OK, but what about code blocks then? These can span several lines, and the temporary placeholder will only be one line long, right? Similarly with CriticMarkup: it can do a multiline capture with a single-line (temporary) replacement. I might still not be understanding the issue correctly, though. 😄

dometto:
> OK, but what about code blocks then? These can span several lines, and the temporary placeholder will only be one line long, right? Similarly with CriticMarkup: it can do a multiline capture with a single-line (temporary) replacement. I might still not be understanding the issue correctly, though.

No, that's exactly the issue! In order for this strategy to work, we'd have to enforce that the placeholders are as long (in lines) as the snippet they're placeholding for. :) So with the current filters that wouldn't work. I was mostly wondering how doable it would be to enforce this in a clean way. 

Maybe we could have a helper method on the `Filter` class that generates a placeholder for a given snippet, rather than hardcoding the placeholder in each filter's `extract` method??

bartkamphorst:
> Maybe we could have a helper method on the Filter class that generates a placeholder for a given snippet, rather than hardcoding the placeholder in each filter's extract method?

Yes, but this could still easily go wrong with user-provided filters who don't (yet) use the helper. Or what about a filter that reduces superfluous whitespace? That could inadvertently eat up some newlines that the helper has inserted. In short, I'm not sure how viable this would be. As an alternative, isn't the original text exposed to the filters by any chance?

dometto:
It wouldn't be hard to give the filters access to the original source (although it's not very efficient). How would that allow us to get a robust sourcemap?

bartkamphorst:
How about a SourceMap filter that sits at the very beginning of the chain, does nothing for `extract` but uses `process` to generate the sourcemap? It would have access to the processed data. Would that work?

dometto:
I thought of that, but the problem is that after some `extract` calls have been done, the markup looks very different from its state before the filters were applied. So the linenumberings for the headers that SourceMap detects doesn't (necessarily) correspond to the linenumbers in the original document. And the source the user is presented with in the editor is the markup in its unaltered, pre-filter shape.

One experimental approach I'm trying is this:
* Before executing any filters in the chain, render the document and generate a sourcemap.
* Execute the pre-render filters
* Execute the render filter, generating a new sourcemap
* Compare the headers found in the first and second sourcemaps, based on their anchor names
  * any headers found in the original one that *aren't* in the latter one should go (because they were apparently false positives: e.g., a header tag embedded in a code or yaml block).
  * However, we should use the linenumbers found in the original sourcemap

Then we must handle edge cases. Consider the following markdown:
```markdown
---
bla
# First header
---

# First header
```

The original sourcemap will contain:
```ruby
{'first-header': 3, 'first-header-1': 6}
``` 
The second sourcemap will contain:
```ruby
{'first-header': 2}
```
Because the YAML has been cut off, it finds `first-header` at line 2 of the (modified) document. However, `first-header` in the rendered document should be pointing to line number 6 in the original source. So we must set the line number for `first-header` (in the second sourcemap) to the value for `first-header-1` (in the original sourcemap).

Downside to this entire method: requires an extra rendering step. And it's complicated..

bartkamphorst:
> And it's complicated..

I agree. It is complicated and computationally expensive. Since gollum already suffers from slow page loads, I'm not sure we should add additional rendering unless the trade-off is absolutely worth it.

dometto:
So here is my last attempt at this using JS. It utilizes ace's syntax highlighting to identify markdown header 'tokens' (in ace-speak). This means header markers in codeblocks etc. are correctly ignored.

In `gollum.editor.js`:

```javascript
    var session = window.ace_editor.session;
    var num_lines = session.getLength();
    var sourcemap = new Object;
    for (i = 0; i < num_lines; i++) {
      token = session.getTokenAt(i,0);
      if (token != null && token['type'].startsWith('markup.heading.')) {
        if (token['value'].startsWith('#')) {
          line = i;
          header = session.getLine(line).substring(token['value'].length);
        } else { // It's a ==== or ---- style header, header text is on the line below
          line = i-1;
          header = session.getLine(line);
        }
        header = header.trim();
        header = header.replace(/[^a-z0-9_\s]/gi, ' ')
        header = header.replace(/\s\s+/g, ' ');
        header = header.split(/\s/).join('-').toLowerCase();
        if (sourcemap[header] == undefined) {
          sourcemap[header] = line; 
        } else {
          var index = 1;
          while (sourcemap[header + '-' + index] != undefined) {
            index++;
          }
          sourcemap[header + '-' + index] = line;
        }
      }
    }
    console.log(sourcemap);
```
This works only for markdown, but we could define functions such as these on a per-language basis in the language definitions. If no sourcemap function is defined, edit sections functionality is not supported.

The above snippet correctly produced a sourcemap for my markdown document, where the keys are equal to the anchor names that gollum produces, so we can easily match them up with whatever section the user wanted to edit:

```
Object { "bilbo-baggins-is-home-md-hobbit": 0, hello: 13, "hello-1": 16, test: 19 } # 'hello-1' correctly generated

```

Edge cases:
* this is based on ace's markdown language definition, so header elements in code blocks etc. are ignored. By gollum also supports e.g. plantuml block-level syntax elements (plantuml is the only one I've identified till now), if you add `# Header` in there ace will consider it as a header. 
  * possible to work around this by extending ace's markdown language definition to include gollum-specific syntax, which is nice anyway.
* `# Link in a [[Home.md header]]` will be rendered in gollum as `Link in a header` (anchor: `#link-in-a-header`), but Ace doesn't know about gollum link tags, so it will parse that as `Link in a Home.md header` (anchor: `link-in-a-home-md-header`).
* Similarly for Macros

I don't know how to deal with the last two edge cases. We could:
* Just bork and not produce the sourcemap if we see Macros or link tags in any of the headers
* Use Ace to collect all the header tokens and do an Ajax call to gollum, letting gollum render them all, and thus we're guaranteed to get back exactly the anchors that gollum would produce.
  * requires a partial/fragment route to be added to the app

@bartkamphorst what do you think? Is this viable? If not, I think we should give up on editable sections -- or at leasrt I don't see how it could work

bartkamphorst:
> Is this viable? If not, I think we should give up on editable sections -- or at leasrt I don't see how it could work.

I think this is the most viable solution so far. I don't know how to deal with the edge cases either, but then again I also don't really understand how the edge cases would affect the functionality. Is that gollum would show the wrong section? Would it still be "near enough" the intended section?

dometto:
> # Link in a [[Home.md header]] will be rendered in gollum as Link in a header (anchor: #link-in-a-header), but Ace doesn't know about gollum link tags, so it will parse that as Link in a Home.md header (anchor: link-in-a-home-md-header).

Just realized this won't be a problem, as we wisely decided not to add the `editable` class to a header if its content contains a tag placeholder:

https://github.com/gollum/gollum-lib/commit/cde561e06c62fcfe1e2146e672b6c66a2fe4d28d#diff-2645eea806fd1f71019badc6db97ac15R28

i.e. headers with links or other markup in them won't be clickable in the first place.

That leaves the issue of custom block level syntax (UML for sure --- anything else? custom code blocks?), for which this still applies:

> possible to work around this by extending ace's markdown language definition to include gollum-specific syntax, which is nice anyway.

PR #312, opened 2018-11-16
Remove useless code from markup.rb

dometto:
This looks like a remainder of days long gone. It says it's expecting `data` to be the result of the rendering operation, but the filter chain hasn't even been executed yet. Moreover, the `render` method doesn't take an optional block, so we'll never actually yield. Probably someone forgot to remove this piece of code when the actual execution of the filter chain was moved to the `process_chain` method. Right now it's not doing anything.

### DIFF
--- a/lib/gollum-lib/markup.rb
+++ b/lib/gollum-lib/markup.rb
@@ -123,6 +123,11 @@ module Gollum
         data = filter.extract(data)
       end
 
+      # Since the last 'extract' action in our chain *should* be the markup
+      # to HTML converter, we now have HTML which we can parse and yield, for
+      # anyone who wants it
+      yield Nokogiri::HTML::DocumentFragment.parse(data) if block_given?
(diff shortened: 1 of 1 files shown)


dometto:
I stand corrected: the `render` method is actually taking a block, and the tests expect it to. Nevertheless, the doc that is yielded is based on the unprocessed page data. So we should either:
* remove the tests that expect something to be yielded here
* pass the block, if given, on to `process_chain`, and move this code to that method (right after the extract operation is completed)

A propos, this commit seems to be the one that accidentally left this code in place:
https://github.com/gollum/gollum-lib/commit/542c1f0777b655f7c1b40fee66a202c963afdd00

bartkamphorst:
I was looking at this code today too, wondering about its purpose. :) I think your second solution of executing the block after the process_chain is probably the best way to go.

PR #313, opened 2018-11-18
Do not include ancestry in anchors. Resolves #167

### LINKED ISSUE #167: Crop large anchor ID's
This is an example of a generated anchor ID on the wiki I'm working on:

`Manual/prefs#preferences_editor-preferences_smart-editing_configuring-word-or-character-wrap`

It's veeery long, too long. Could this be handled more intelligently? eg. it seems to be including all parent headers in the anchor, this seems unnecessary. Additionally it would be nice if we could force a max length for these anchors, and have it crop if the max length is reached. 

I understand the main reason for this is to ensure the anchor is unique, but you could also generate a hash based on the current information used and append this.

Far as I'm aware these hashes don't affect SEO, so there should be no concerns there.

dometto:
#167 rightly complains that anchors get too long because the full header-ancestry is included in them (so end up e.g. with `Manual/prefs#preferences_editor-preferences_smart-editing_configuring-word-or-character-wrap`). https://github.com/gollum/gollum/issues/1261 complains that github and gollum use different schemes to generate anchors.

This PR solves the first issue by conforming the way we generate anchors to GitHub's scheme:
```
# Ecthelion

## Boromir

## Faramir

# Boromir

## Ecthelion
```
...generates these anchors:
```
#ecthelion
#boromir
#faramir
#boromir-1
#ecthelion-1 
```

As an added advantage, this somewhat simplifies the logic in the TOC filter.

- [ ] Update release notes

### DIFF
--- a/lib/gollum-lib/filter/toc.rb
+++ b/lib/gollum-lib/filter/toc.rb
@@ -70,7 +70,7 @@ class Gollum::Filter::TOC < Gollum::Filter
   # Prefixes duplicate anchors with an index
   def generate_anchor_name(header)
     name = header.content
-    level = header.name.gsub(/[hH]/, '').to_i
+    level = header.name[1..-1].to_i
 
     # normalize the header name
     name.gsub!(/[^\d\w\u00C0-\u1FFF\u2C00-\uD7FF]/, "-")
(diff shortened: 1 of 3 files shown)


lib/gollum-lib/filter/toc.rb
```diff
@@ -70,7 +70,7 @@ def process(data)
   # Prefixes duplicate anchors with an index
   def generate_anchor_name(header)
     name = header.content
-    level = header.name.gsub(/[hH]/, '').to_i
+    level = header.name[1..-1].to_i
```
bartkamphorst: Just for my understanding: what is this doing? Can you provide an example?

bartkamphorst:
Looks clean. Nice work! 👍

lib/gollum-lib/filter/toc.rb
dometto: `header.name` is e.g. `h2` (set by nokogiri). The gsub was removing the h character in order to get the header level. `[1..-1]` does the same but in a more efficient way:

```ruby
2.4.1 :001 > 'h12'[1..-1]
 => "12" 
2.4.1 :002 > 'H3'[1..-1]
 => "3" 
```

petrkalina:
I have the same issue - i.e. that github anchors are generated different than gollum.but I'm running the latest v. 4 release installed via

```
RUN gem install gollum
RUN gem install github-markdown
```

is there any solution for

PR #315, opened 2018-11-29
Render pure anchor page links. Resolves gollum#1236

dometto:
Issue: https://github.com/gollum/gollum/issues/1236

This PR resolves the issue by catching the anchor-only edge case in the tag filter. Additionally, an exception had to be made to the sanitization rules: the sanitizer currently prefixes 'wiki-' to all href elements that start with '#'. I think this is in order for gollum's css to apply to elements generated in the filter chain, and the exception should thus not pose a security risk -- but please see if that makes sense.

Flow control was a little complicated, so please let me know if you see better options.

### DIFF
--- a/lib/gollum-lib/filter/tags.rb
+++ b/lib/gollum-lib/filter/tags.rb
@@ -215,7 +215,7 @@ class Gollum::Filter::Tags < Gollum::Filter
     page      = find_page_from_path(link)
 
     # If no match yet, try finding page with anchor removed
-    unless page
+    if page.nil?
       if pos = link.rindex('#')
         extra = link[pos..-1]
         link  = link[0...pos]
@@ -223,7 +223,7 @@ class Gollum::Filter::Tags < Gollum::Filter
         extra = nil
       end
 
-      if link.empty? && !extra.nil? # Internal anchor link, don't search for the page but return immediately
+      if link.empty? && extra # Internal anchor link, don't search for the page but return immediately
         return generate_link(nil, pretty_name, extra, :internal_anchor)
       end
 
@@ -273,11 +273,8 @@ class Gollum::Filter::Tags < Gollum::Filter
   #
   # Returns a String href.
   def generate_href_for_path(path, extra = nil)
-    if path.nil? && !extra.nil? # Internal anchor link
-      extra
-    else
-      "#{trim_leading_slash(::File.join(@markup.wiki.base_path, path))}#{extra}"
-    end
+    return extra if !path && extra # Internal anchor link
+    "#{trim_leading_slash(::File.join(@markup.wiki.base_path, path))}#{extra}"
   end
 
   # Construct a CSS class and attribute string for different kinds of links: internal Pages (absent or present) and Files, and External links.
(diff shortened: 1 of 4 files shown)


lib/gollum-lib/filter/tags.rb
```diff
@@ -278,6 +291,8 @@ def css_options_for_link(kind)
       'class="internal absent"'
     when :page_present
       'class="internal present"'
+    when :internal_anchor
+      'class="internal-anchor"'
```
dometto: I had `internal anchor` first, but it turns out the `anchor` class is already in use and actually makes links invisible. :)

dometto: To clarify, the class is not defined in the CSS at present, but this is to a) make sure we can add it in the future b) make the exception in `sanitization.rb`

lib/gollum-lib/filter/tags.rb
```diff
@@ -215,12 +215,21 @@ def process_page_link_tag(link_part, pretty_name = nil)
     page      = find_page_from_path(link)
 
     # If no match yet, try finding page with anchor removed
-    if (page.nil? && pos = link.rindex('#'))
-      extra     = link[pos..-1]
-      link      = link[0...pos]
-      page      = find_page_from_path(link)
+    unless page
```
bartkamphorst: I prefer `if page.nil?`.

lib/gollum-lib/filter/tags.rb
bartkamphorst: Wouldn't it be cleaner to add `internal anchor` CSS as a subclass of anchor?

test/test_markup.rb
```diff
@@ -266,6 +266,21 @@
     assert_html_equal "<p>\na <a class=\"internal present\" href=\"/Potato.mediawiki\">Potato Heaad</a> </p>", output
   end
 
+  test "page link with anchor only" do
```
bartkamphorst: The test names do not make it clear that they concern _internal_ anchors.

dometto:
> See my comment in the main conversation about this.

@bartkamphorst did I miss this comment? Or was this just about adding `internal anchor` to the CSS? Another way to go would be to rename the class to `internal page` (as in, the link target is internal to the wiki, and to be more precise, it's on the same page).

dometto:
I've tried playing around with the CSS to get the desired results when using the class `internal anchor`, but I can't figure it out. The classes are cumulative, so I can't easily negate the results of the `anchor` class is a neat way (I can set every property that's set in the `anchor` class to it's default for normal links, but that doesn't feel very clean). What shall we do?

bartkamphorst:
> I can set every property that's set in the anchor class to it's default for normal links, but that doesn't feel very clean

Overriding the parent class is the standard way to do this though. However, looking at it again, I've changed my mind and think `internal-anchor` works! So leave it as is. 😌

bartkamphorst:
As an alternative, you could also use `internal anchorlink`.

dometto:
Remaining test failure seems to be due to the placing of href and id attributes being reversed in nokogiri on the latest jruby on Travis. Not sure how we can make the test more lenient.

PR #319, opened 2019-01-03
Feature/fix gem 3

scolytus:
While building gollum in my Docker environment the build started to fail. The Ruby 2.5 images now contains gem version 3. There the already deprecated options `--no-rdoc` and `--no-ri` were removed.


bartkamphorst:
@scolytus Same question here: could you squash the two commits into one?

PR #322, opened 2019-02-16
Refactor find

dometto:
Radically simplifies the page-lookup API. Rather than `paged` and `page` with complicated method signatures, we now just have `Wiki#page(path, version)`. The main consequence: until now it was possibly to find `Foo/Bar.md` by doing `wiki.paged('Bar.md')`. In effect this meant that the result of the page lookup was not uniquely determined. This no longer works: you *have* to specify the subpath, so the result is uniquely determined. I had to modify lots of tests to conform to the new API.

I have also made lookup of Files case sensitive (like Pages).

Moreover, with a view to the upcoming solution for https://github.com/gollum/gollum/issues/1052 , this also:

* Makes sure Page/File lookups prepend the page_file_dir to the queried path.
* Modifies the generation of url paths for pages and files: it chops off the page_file_dir


lib/gollum-lib/file.rb
```diff
@@ -19,13 +19,15 @@ def initialize(wiki)
...
-        path << name
-      end
-      path
+      path = if self.path.include?('/')
+        self.path.sub(/\/[^\/]+$/, '/')
+          else
+            ''
+          end
+      path = path[@wiki.page_file_dir.length+1..-1] if @wiki.page_file_dir # Chop off the page file dir plus the first slash if necessary 
+      path << name
```
dometto: Not sure the `<< name` here is appropriate, as it looks like the `name` method sometimes returns the full `path` itself! If it's a mistake, this was already broken. Help appreciated.

dometto: Given this issue and substantial overlap, I think we should consider making `File` a subclass of `Page`, or having some other class hierarchy that avoids duplication.

bartkamphorst: > having some other class hierarchy that avoids duplication.

Yes, that makes sense. I would think that `Gollum::Page` would be a special case of the more general `Gollum::File` class, but that may just be a personal preference.

lib/gollum-lib/file.rb
```diff
@@ -19,13 +19,15 @@ def initialize(wiki)
...
     # Returns the String url_path
+
     def url_path
-      path = self.path
-      if path.include?('/')
-        path = path.sub(/\/[^\/]+$/, '/')
-        path << name
-      end
-      path
+      path = if self.path.include?('/')
```
bartkamphorst: I'm not a huge fan of this kind of assignment to be honest, since it requires a substantial look ahead to understand how the assignment is determined.

dometto: Agreed

lib/gollum-lib/file.rb
bartkamphorst: > Not sure the `<< name` here is appropriate

This was introduced [here](https://github.com/gollum/gollum-lib/commit/b10a0c98ba12a5ed8f8a05ff69d5ca68f1b65ad9) to let `File.url_path` behave in the same way as `Page.url_path`. Another reason for creating a superclass I would say.

lib/gollum-lib/file.rb
```diff
@@ -145,15 +147,21 @@ def get_disk_reference(name, commit)
...
     # for which on_disk? is actually true.
     def find(name, version, try_on_disk = false)
-      checked = name.downcase
       map     = @wiki.tree_map_for(version)
+      if @wiki.page_file_dir
+       path = ::File.join(@wiki.page_file_dir, name)
+      else
+       path = name
+      end
+      query = ::File.join('/', path)
```
bartkamphorst: Perhaps we could define a constant `PATH_SEPARATOR='/'` and use that throughout the code base instead of sprinking `'/'` where needed?

lib/gollum-lib/file.rb
```diff
@@ -145,15 +147,21 @@ def get_disk_reference(name, commit)
...
+
       commit  = version.is_a?(Gollum::Git::Commit) ? version : @wiki.commit_for(version)
 
-      if (result = map.detect { |entry| entry.path.downcase == checked })
-        @path    = name
+      if (result = map.detect { |entry| ::File.join('/', entry.path) == query })
```
bartkamphorst: This seems like a costly operation to do for all entries. Is there an alternative?

lib/gollum-lib/file.rb
bartkamphorst: > as it looks like the `name` method sometimes returns the full `path` itself

Only in the sense that the "full path" sometimes just _is_ the name of the page (any page at root level).

test/test_markup.rb
```diff
@@ -666,7 +666,7 @@
     index.add("greek/Bilbo-Baggins.md", "a [[Alpha|alpha.jpg]] b")
     index.commit("Add alpha.jpg")
 
-    page   = @wiki.page("Bilbo-Baggins")
+    page   = @wiki.page("greek/Bilbo-Baggins")
```
bartkamphorst: These are relative paths; should I read them as being relative to root though?

test/test_page.rb
```diff
@@ -56,7 +60,7 @@
   end
 
   test "nested url_path" do
-    page = @wiki.page('Eye-Of-Sauron')
+    page = @wiki.page('/Mordor/Eye-Of-Sauron')
```
bartkamphorst: Here you use an absolute instead of relative path. Does it matter? Should we be consistent in our use?

lib/gollum-lib/file.rb
dometto: There is ::File::SEPARATOR, which I'll use!

test/test_markup.rb
dometto: Yes, `./greek` in this case is the page file dir. Is that intuitive?

test/test_page.rb
dometto: I was thinking it would be intuitive to not require the leading slash for a path to start with root, but let it also work with an explicit slash. Is that ok?

lib/gollum-lib/file.rb
dometto: Would a regexp comparison be cheaper?

lib/gollum-lib/file.rb
bartkamphorst: Perfect. 👍

lib/gollum-lib/file.rb
bartkamphorst: What would the regexp be? I will run a quick benchmark.

test/test_markup.rb
bartkamphorst: > Yes, ./greek in this case is the page file dir. Is that intuitive?

Not to me. 😬 If `page_file_dir` is a system-wide setting set to `greek`, shouldn't it be assumed that when you call `@wiki.page("Bilbo-Baggins")` that you want `greek/Bilbo-Baggins`?

test/test_page.rb
bartkamphorst: Yes, but it would be good to add a note in the tests to explain this and a remark in the release notes as well.

test/test_markup.rb
dometto: Excuse me, you're right of course. `'greek'` in this example is not the page file dir. It is just a relative path that you should, indeed, read as relative to root. See below.

lib/gollum-lib/file.rb
dometto: `/\/+#{entry.path}/ =~ query`

dometto:
@bartkamphorst please check this PR again. I've refactored the `Page` class as a subclass of the `File` class. The only bit of the `find` mechanism that is different for `Page` and `File` now is implemented in the private `path_match` method (which both classes need to have implemented).

lib/gollum-lib/file.rb
bartkamphorst:
```ruby
map = ["a", "/b", "/c/d"] 
query = '/b'
puts Benchmark.measure{
  50_000.times do
    map.detect{|entry| ::File.join('/', entry) == query}
  end
}
#=>   0.100000   0.010000   0.110000 (  0.112311)

puts Benchmark.measure{
  50_000.times do
    map.detect{|entry| /\/+#{entry}/ =~ query}
  end
}
#=>  0.920000   0.020000   0.940000 (  0.941059)
```

If this benchmark makes sense (@dometto ?) then straightforward string comparison is much faster than regexp and you should keep the current solution.

lib/gollum-lib/page.rb
```diff
@@ -447,13 +346,11 @@ def inspect
...
     private
 
-    def construct_path(name)
-      path = if self.path.include?('/')
-        self.path.sub(/\/[^\/]+$/, '/')
-          else
-            ''
-          end
-      path << name
+    def path_match(query, entry)
```
bartkamphorst: how about:
```ruby
def path_match(query, entry)
  return false unless self.class_valid_extension?(entry.name)
  entry_name = self.class.valid_extension?(query) ? entry.name : ::Gollum::Page.strip_filename(entry.name)
  query == ::File.join(::File::SEPARATOR, entry.dir, entry_name)
end

**Comment by bartkamphorst** [comment]
>  I've refactored the Page class as a subclass of the File class.

Sweet! There's still a few unresolved discussion points, but it's shaping up nicely!

File: test/test_markup.rb (hunk 5 above)
Comment (bartkamphorst) [reply test/test_markup.rb]: OK.

File: lib/gollum-lib/file.rb (hunk 4 above)
Comment (dometto) [reply lib/gollum-lib/file.rb]: Thanks! Great to actually have some data here. I think the results also make sense, as File::Join appears to be implemented entirely in C.

File: lib/gollum-lib/page.rb (hunk 7 above)
Comment (dometto) [reply lib/gollum-lib/page.rb]: Yes! Thanks

File: lib/gollum-lib/file.rb (hunk 3 above)
Comment (dometto) [reply lib/gollum-lib/file.rb]: Hmm, I'm not sure now what the right approach is. Seeing as the generated paths will end up being used as the page or file's `url_path`, we really don't want anything other than slashes (backslashes?) popping up there. According to [this](https://www.ruby-forum.com/t/path-separator-and-windows/51339), slash should work even on Windows. I've reinstated the slashes for now, but let me know what you think is the best approach!

File: test/test_page.rb (hunk 6 above)
Comment (dometto) [reply test/test_page.rb]: 👍

**Comment by dometto** [comment]
I realized it was kind of strange that `#find` was an instance method on `Page` and `File`. Calls to it were of the form `Page.new(wiki).find(path, version)`. That made it look as if one could have an adequately instantiated page without it actually being in the repo.

To remedy this, I've changed `find` to be a class method on `Page` and `File`. So calls are now of the form `page = Page.find(wiki, path, version)`, which seems much more natural. Fortunately, `find` was only actually being called on two lines in the `wiki.rb`, so it was an easy change. :) All other uses should and do interact via `Wiki#page` and `Wiki#file`.

**Comment by bartkamphorst** [comment]
> To remedy this, I've changed find to be a class method on Page and File

👍

**Comment by dometto** [comment]
...and then I realized this allowed a further radical simplification of the API.

Until now the way to do things was e.g.:

```ruby
page = Page.new(wiki)
page.historical = blah
page.version = blah
page.populate # Sets further instance variables on page
```
This meant that you could still have an 'unbaked' Page or File object which was for all intents and purposes unitialized. I've now added a number of required parameters to `#initialize` -- it all happens in #initialize. Benefits:

* no confusing 'unbaked' objects possible
* no more separate calls to `#populate`
* a separate `Page.find` method is no longer required. `#initialize` is the only method that needs redefining in the subclass.

The only reason I could find for the previous way of doing things was that there is a `Wiki#preview_page` method which generates a "mock" Page object for preview purposes, which therefore had to be 'unbaked'. But I've reimplemented this with the new `PreviewPage` subclass of `Page`.

lib/gollum-lib/file.rb
```diff
@@ -2,55 +2,103 @@
...
+    # try_on_disk - If true, try to return just a reference to a file
+    #               that exists on the disk.
+    #
+    # Returns a Gollum::File or nil if the file could not be found. Note
+    # that if you specify try_on_disk=true, you may or may not get a file
+    # for which on_disk? is actually true.
+    def self.find(wiki, path, version, try_on_disk = false)
+      map = wiki.tree_map_for(version.to_s)
+
+      if wiki.page_file_dir
```
bartkamphorst: I think here the ternary operator would do nicely:
```ruby
query_path = wiki.page_file_dir ? ::File.join(wiki.page_file_dir, path) : path
```
Sweet and short, and still understandable!

lib/gollum-lib/file.rb
```diff
@@ -131,36 +155,18 @@ def get_disk_reference(name, commit)
...
-        if try_on_disk && get_disk_reference(name, commit)
-          @on_disk = true
-        else
-          @blob = result.blob(@wiki.repo)
-        end
-
-        self
-      end
+    def construct_path(name)
+      path = if self.path.include?('/')
```
bartkamphorst: Ternary operator would work here as well:
```ruby
path = self.path.include?('/') ? self.path.sub(/\/[^\/]+$/, '/') : ''
```
Still not great, but shorter. 😉

lib/gollum-lib/file.rb
```diff
@@ -131,36 +155,18 @@ def get_disk_reference(name, commit)
...
+        self.path.sub(/\/[^\/]+$/, '/')
+          else
+            ''
+          end
+      path = path[@wiki.page_file_dir.length+1..-1] if @wiki.page_file_dir # Chop off the page file dir plus the first slash if necessary
```
bartkamphorst: I would prefer to have the comment above the relevant line for legibility.

lib/gollum-lib/page.rb
```diff
@@ -1,19 +1,22 @@
...
-    # Parent page if this is a sub page
-    #
-    # Returns a Page
-    attr_accessor :parent_page
+    class << self
+      # For use with self.find: returns true if the given query corresponds to the in-repo path of the BlobEntry. 
+      #
+      # query     - The String path to match.
+      # entry     - The BlobEntry to check against.
+      def path_match(query, entry)
```
bartkamphorst: Lovely code! 😄

lib/gollum-lib/page.rb
```diff
@@ -398,7 +292,7 @@ def populate(blob, path=nil)
     # Returns the String path.
     def tree_path(treemap, tree)
       if (ptree = treemap[tree])
-        tree_path(treemap, ptree) + '/' + tree.name
+        "#{tree_path(treemap, ptree)}#{::File::SEPARATOR}#{tree.name}"
```
bartkamphorst: Much better!

lib/gollum-lib/wiki.rb
```diff
@@ -162,25 +162,13 @@ def exist?
...
     # Public: Get the formatted page for a given page name, version, and dir.
     #
-    # name    - The human or canonical String page name of the wiki page.
+    # path    - The String path to the the wiki page (may or may not include file extension).
     # version - The String version ID to find (default: @ref).
-    # dir     - The directory String relative to the repo.
     #
     # Returns a Gollum::Page or nil if no matching page was found.
-    def page(name, version = @ref, dir = nil, exact = false)
+    def page(path, version = @ref)
```
bartkamphorst: If `@ref` always set?

bartkamphorst: Or can `version` be `nil`?

lib/gollum-lib/wiki.rb
```diff
@@ -194,7 +182,7 @@ def paged(name, dir = nil, exact = false, version = @ref)
     # that if you specify try_on_disk=true, you may or may not get a file
     # for which on_disk? is actually true.
     def file(name, version = @ref, try_on_disk = false)
-      ::Gollum::File.new(self).find(name, version, try_on_disk)
+      ::Gollum::File.find(self, name, version, try_on_disk)
```
bartkamphorst: Nice! 👍

lib/gollum-lib/wiki.rb
```diff
@@ -207,13 +195,7 @@ def file(name, version = @ref, try_on_disk = false)
...
     # Returns the in-memory Gollum::Page.
     def preview_page(name, data, format)
-      page = ::Gollum::Page.new(self)
-      ext  = ::Gollum::Page.format_to_ext(format.to_sym)
-      filename = "#{name}.#{ext}"
-      blob = OpenStruct.new(:name => filename, :data => data, :is_symlink => false)
-      page.populate(blob)
-      page.version = @access.commit(@ref)
-      page
+      ::Gollum::PreviewPage.new(self, "#{name}.#{::Gollum::Page.format_to_ext(format.to_sym)}", data, @access.commit(@ref))
```
bartkamphorst: Clean. The only worry I have is that the method signature for initialization is not completely intuitive, but I do like how short it is.

lib/gollum-lib/file.rb
dometto: Excellent!

lib/gollum-lib/file.rb
dometto: 👍

lib/gollum-lib/wiki.rb
dometto: the `nil` case is handled on the next line

dometto:
Incorporated your last feedback, @bartkamphorst! Do you think the additions to the `README` describe the changes to the API clearly enough?

bartkamphorst:
> Do you think the additions to the README describe the changes to the API clearly enough?

Yes, those are clear. I'll have a look at the other changes tomorrow.

bartkamphorst:
>  I've reimplemented this with the new PreviewPage subclass of Page.

Nice. 👍

lib/gollum-lib/blob_entry.rb
```diff
@@ -47,10 +47,7 @@ def blob(repo)
...
     # Returns a Gollum::Page instance.
     def page(wiki, commit)
-      blob         = self.blob(wiki.repo)
-      page         = ::Gollum::Page.new(wiki).populate(blob, self.dir)
-      page.version = commit
-      page
+      ::Gollum::Page.new(wiki, self.blob(wiki.repo), self.dir, commit)
     end
 
     # Gets a File instance for this blob.
```
bartkamphorst: Just for my understanding: how can one tell if one should call `page` or `file` on a `blob_entry` object?

dometto: I think only based on whether it has a valid page extension. I think these methods on BlobEntry are now only being called from somewhere within wiki.rb. They don't seem essential, but I decided to leave them in for convenience. We could also try to get rid of them entirely if you think that would be less confusing

lib/gollum-lib/file.rb
```diff
@@ -131,36 +151,15 @@ def get_disk_reference(name, commit)
...
-        if try_on_disk && get_disk_reference(name, commit)
-          @on_disk = true
-        else
-          @blob = result.blob(@wiki.repo)
-        end
-
-        self
-      end
+    def construct_path(name)
+      path = self.path.include?('/') ? self.path.sub(/\/[^\/]+$/, '/') : ''
```
bartkamphorst: To better express the intent here, you could put the `sub` call into a private method like so:
```ruby
def get_dirpath_from_filepath(filepath)
  filepath.sub(/\/[^\/]+$/, '/')
end
#...
def construct_path(name)
  path = self.path.include?('/') ? get_dirpath_from_filepath(self.path) : ''
  #...
end
```

And, as a related question, couldn't we just replace the `sub` call altogether with a much more expressive call to `File::dirname`? For example:
```ruby
def get_dirpath_from_filepath(filepath)
  ::File.dirname(filepath) << ::File::SEPARATOR
end
```

lib/gollum-lib/blob_entry.rb
bartkamphorst: What would happen if one were to call `page` on a BlobEntry object that is actually not a page?

bartkamphorst:
No further remarks at this point except the two open conversations.

lib/gollum-lib/blob_entry.rb
dometto: I think you would end up with a Page object that would throw errors (hopefully) when you try to access the rendered_data method, because no renderer exists for the file type in question

lib/gollum-lib/file.rb
dometto: Actually, I just noticed all this was weirdly complicated. `construct_path` was chopping off the filename from `self.path` in order to... then add the filename to it. I managed to get rid off the helper method entirely, and changed the `url_path` method to:
```ruby
    # Public: The url path required to reach this file within the repo.
    #
    # Returns the String url_path
    def url_path
      # Chop off the page_file_dir and first slash if necessary
      @wiki.page_file_dir ? self.path[@wiki.page_file_dir.length+1..-1] : self.path
    end
```

bartkamphorst: Erm, interesting... Do the tests adequately reflect all the cases (pages at root, pages at subdir level, uploaded files, etc.) so that we can be certain that this works properly across the board?

lib/gollum-lib/blob_entry.rb
bartkamphorst: > I think only based on whether it has a valid page extension

Wouldn't it be preferable to hide this logic from the user? If there is no use case for trying to cast a BlobEntry into a Page while it is actually a File (I can't think of one), then why not let Gollum handle this internally? For example, `BlobEntry#contents` could return either a Gollum::Page (which is a subclass of Gollum::File) or a Gollum::File, depending on which one is fitting. Thoughts?

lib/gollum-lib/file.rb
dometto: Good question. I outputted the url_paths generated in the tests, and there is a nice variety of them, fortunately. I now added tests for url_path when page_file_dir is set. Everything seems to work fine.

lib/gollum-lib/blob_entry.rb
dometto: Yes I was wondering about that too. We could have `to_page_or_file` method (better name?), but in all contexts where we'd use that, we actually already know whether it's going to be a page or file. [Here](https://github.com/gollum/gollum-lib/blob/gollum-lib-5.x/lib/gollum-lib/wiki.rb#L683-L684), for example, we'd still have to do the `valid_page_name?` check to exclude subpages (which have a valid page extension, but are not a valid standalone page). Do you still think it would be better to have the generic method?

bartkamphorst: Okay. Leave it as is for now. 👍

PR #326, opened 2019-04-25
Implement Wiki#write_file

dometto:
Adds `Wiki#write_file` to the API.

Until now, the API only allowed for adding *pages* to the wiki, and not *files*. That is why the logic for uploading files in gollum is [directly doing commits from the frontend](https://github.com/gollum/gollum/blob/master/lib/gollum/app.rb#L210-L218). Ouch.

This PR is a prerequisite for merging https://github.com/gollum/gollum/pull/1369 as it ensures that the `:page_file_dir` option is respected when writing a new file to the wiki.

While refactoring the `Committer` class for this PR, I also noticed that we were still throwing `DuplicatePageError`s when the updated path case-**in**sensitively matched a file in the wiki. Since we have case sensitive filepaths in `5.x`, that's a bug. I added a test to guard against that. This also makes the `Committer#add_to_index` method much cleaner and more efficient.

### DIFF
--- a/lib/gollum-lib/wiki.rb
+++ b/lib/gollum-lib/wiki.rb
@@ -1,4 +1,6 @@
 # ~*~ encoding: utf-8 ~*~
+require 'pathname'
+
 module Gollum
   class Wiki
     include Pagination
@@ -125,6 +127,7 @@ module Gollum
       @path                 = path
       @repo_is_bare         = options.fetch :repo_is_bare, nil
       @page_file_dir        = options.fetch :page_file_dir, nil
+      @page_file_dir        = Pathname.new("/#{@page_file_dir}").cleanpath.to_s[1..-1] if @page_file_dir
       @access               = options.fetch :access, GitAccess.new(path, @page_file_dir, @repo_is_bare)
       @base_path            = options.fetch :base_path, "/"
       @repo                 = @access.repo
@@ -802,9 +805,14 @@ module Gollum
     #
     # Returns a String path.  .
     def merge_path_elements(dir, name, format)
-      result = ::File.join([@page_file_dir, '/', dir, self.page_file_name(name, format)].compact)
-      raise Gollum::IllegalDirectoryPath if @page_file_dir && !Pathname.new(result).cleanpath.to_s.start_with?(@page_file_dir)
-      result[0] == '/' ? result[1..-1] : result
+      result = ::File.join([@page_file_dir, dir, self.page_file_name(name, format)].compact)
+      result = Pathname.new(result).cleanpath.to_s
+      if @page_file_dir
+        raise Gollum::IllegalDirectoryPath unless result.start_with?("#{@page_file_dir}/")
+        result
+      else
+        result[0] == '/' ? result[1..-1] : result
+      end
     end
 
     def write(path, data, commit = {})
--- a/lib/gollum-lib/wiki.rb
+++ b/lib/gollum-lib/wiki.rb
@@ -236,7 +236,7 @@ module Gollum
     # Returns the String SHA1 of the newly written version, or the
     # Gollum::Committer instance if this is part of a batch update.
     def write_page(name, format, data, commit = {}, dir = '')
-      write_page_or_file(merge_page_elements(dir, name, format), data, commit)
+      write_page_or_file(merge_path_elements(dir, name, format), data, commit)
     end
 
     # Public: Write a new version of a file to the Gollum repo root.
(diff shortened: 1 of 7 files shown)


dometto:
To explain the changes to `wiki.rb` more fully:
* Until now, `Committer#add_to_index` was only usable for pages, as it expected the parameters `dir`, `name`, and `format`.
* To make it possible to add files (as opposed to pages) to the index, I changed `#add_to_index` to expect the full `path` as a parameter.
* That then made it necessary to change the other places in `wiki.rb` where `Committer#add_to_index` was being called, and calculate the full `path` beforehand.
  * In order to cleanly calculate the full `path` and respect `@page_file_dir`, I added the new `private` methods to the `Wiki` class.
  * `Wiki#update_file` is a bit of an exception: since in that method we already have an *existing* page, we shouldn't generate the path for `Committer#add_to_index` using the new `merge_page_elements` helper (that would **again** prefix `@page_file_dir` to the page's path)

lib/gollum-lib/wiki.rb
```diff
@@ -810,19 +809,18 @@ def inspect
...
-      raise Gollum::IllegalDirectoryPath if @page_file_dir && !Pathname.new(sanitized).cleanpath.to_s.start_with?(@page_file_dir)
-      sanitized
-    end
-
-    def merge_page_elements(dir, name, format)
-      result = ::File.join(*[sanitized_dir(dir), self.page_file_name(name, format)])
+    # Returns a String path.  .
+    def merge_path_elements(dir, name, format)
+      result = ::File.join([@page_file_dir, '/', dir, self.page_file_name(name, format)].compact)
+      raise Gollum::IllegalDirectoryPath if @page_file_dir && !Pathname.new(result).cleanpath.to_s.start_with?(@page_file_dir)
```
dometto: It feels a bit weird to throw this error in this helper method. What would be a better place, you think? In `write_page_or_file`? Downside of that is other methods, like `delete_file`, wouldn't benefit from the protection against directory traversal. So maybe in `Committer#add_to_index`? That also seems strange..

dometto:
FYI: I just noticed that it *was* actually already possible to write files, and not just pages, to the wiki, because [this](https://github.com/gollum/gollum-lib/blob/gollum-lib-5.x/lib/gollum-lib/wiki.rb#L671) call to `Page.format_to_ext` is actually not validating that `format` is a valid `Page` format. I was confused because `Wiki#page_file_name` seemed to imply that it would generate only valid page names.

However, it still makes sense to me to have a separate `write_file` method, since it is completely untransparent that `write_page` can be used to write a File. Do you think we should lock down the `write_page` method so that it does not allow writing 'mere' files? Or just let both `write_page` and `write_file` exist as different ways of achieving the same thing?

dometto:
> FYI: I just noticed that it was actually already possible to write files, and not just pages, to the wiki, because this call to Page.format_to_ext is actually not validating that format is a valid Page format. I was confused because Wiki#page_file_name seemed to imply that it would generate only valid page names.

Do you also agree that the changes to `Committer` make sense anyway? I thought it was weird that `Committer` was doing work to conjoin `name`, `format`, and `dir`.

dometto:
In the current implementation, `write_page('myblob',...)` called with `nil` as the `format` argument is equivalent to calling `write_file('myblob.js', ...)`. So another idea is to make `write_file` nothing more than a wrapper around `write_page`:
```ruby
    def write_file(name, data, commit = {}, dir = '')
      write_page(name, nil, data, commit, dir)
    end
```
...and folding the logic of `write_page_or_file` all into `write_page`.

bartkamphorst:
> another idea is to make write_file nothing more than a wrapper around write_page

Conceptually, I think it would be nice to have the bulk of the code live in a (private) method called `write` and have `write_page` and `write_file` both be wrappers that call `write` with the appropriate arguments.

dometto:
> Conceptually, I think it would be nice to have the bulk of the code live in a (private) method called write and have write_page and write_file both be wrappers that call write with the appropriate arguments.

👍 

I've now also taken the slightly more radical step of removing the separate `dir` parameter from the `write_page` and `write_file` methods. Calling `write_page('subdirectory/page', ...)` already had the same effect as `write_page('page', ..., 'subdirectory'). The former call seems much more intuitive to me, and it's in line with removing the separate `dir` parameter from `Wiki#page` in [this PR](https://github.com/gollum/gollum-lib/pull/322/files). I've looked and we only need to change one line in `gollum` to conform to the new method signature.

Could you let me know what you think about raising the `IllegealDirectoryPath` exception in the private `merge_path_elements` method?

dometto:
> I've now also taken the slightly more radical step of removing the separate dir parameter from the write_page and write_file methods.

I don't feel very strongly about this, btw, so if you think it makes more sense to keep the separate `dir` parameter, let me know. If you prefer the new situation, I'll update the README and inline docs.

bartkamphorst:
> Could you let me know what you think about raising the IllegealDirectoryPath exception in the private merge_path_elements method?

Seems okay to me.

PR #327, opened 2019-04-28
Fix Page#url_title: no file extension

dometto:
Just noticed that removing the `construct_path` private method from the `Page` class resulted in the `url_path_title` being different from what gollum expects: it was expecting the title without the file extension. This PR removes the file extension from the url_path_title is a straightforward manner, without the need for something like `construct_path`.

Question: is the desired behavior for the title to include the directory path? That was the point of `construct_path`, but won't that result in super long titles? I can remember there being an issue about that, but can't figure out what we decided now.

### DIFF
--- a/lib/gollum-lib/page.rb
+++ b/lib/gollum-lib/page.rb
@@ -116,7 +116,7 @@ module Gollum
     #
     # Returns the String title
     def url_path_title
-      metadata_title || ::File.join(::File.dirname(url_title), name)
+      metadata_title || name
     end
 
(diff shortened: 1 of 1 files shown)


bartkamphorst:
> is the desired behavior for the title to include the directory path?

It used to be, but for `5.x` we landed on the name of the page for the title with a breadcrumb to show the full path (by default, unless overruled by h1-title or metadata title).

dometto:
Ah right, thanks! This should be better then.

PR #330, opened 2019-05-03
Fix deletion with page_file_dir

dometto:
Turns out deleting files/pages was not being tested with the page_file_dir option, and was broken. This PR addresses the issue, adds a test, and refactors the deletion methods.


bartkamphorst:
Looks good! 👍

PR #331, opened 2019-05-05
Chdir to page's directory before rendering. Resolves #170

### LINKED ISSUE #170: AsciiDoc link/include paths are relative from repo root
Because all pages are rendered from the root of the git repository Gollum is serving from, AsciiDoc relative link and include paths don't function as they should.

An easy fix is to wrap [this line](https://github.com/gollum/gollum-lib/blob/master/lib/gollum-lib/filter/render.rb#L6) in a chdir block like this:

``` ruby
Dir.chdir(@markup.dir) do
  data = GitHub::Markup.render(@markup.name, data)
end
```

I'm not sure if that would break any other markup language's rendering, though. Thoughts on whether this should be implemented for all languages, just AsciiDoc, or not at all?

dometto:


lib/gollum-lib/filter/render.rb
```diff
@@ -3,7 +3,11 @@
 class Gollum::Filter::Render < Gollum::Filter
   def extract(data)
     begin
-      data = GitHub::Markup.render(@markup.name, data)
+      working_dir = ::File.join(@markup.wiki.path, @markup.dir)
```
bartkamphorst: For consistency, perhaps a `pathname` solution would be better here?

dometto: 👍

PR #332, opened 2019-07-17
Determine markup format from format name rather than filename

dmyates:
Well, I didn't get to it as soon as I would have liked, but I've redone https://github.com/gollum/gollum-lib/pull/285. To recap, this change makes Gollum use the markup defined in `Gollum::Markup.formats` to render text rather that the source's file extension. This allows users to change which extensions are used for which markups, as I do in [this blog post](https://davidyat.es/2017/09/01/vimwiki-plus-gollum/).

### DIFF
--- a/lib/gollum-lib/filter/render.rb
+++ b/lib/gollum-lib/filter/render.rb
@@ -6,7 +6,7 @@ class Gollum::Filter::Render < Gollum::Filter
       working_dir = Pathname.new(@markup.wiki.path).join(@markup.dir)
       working_dir = working_dir.exist? ? working_dir.to_s : '.'
       Dir.chdir(working_dir) do
-        data = GitHub::Markup.render_s(Gollum::Markup.formats[@markup.format][:name].downcase.to_sym, data)
+        data = GitHub::Markup.render_s(@markup.format, data)
       end
       if data.nil?
         raise "There was an error converting #{@markup.name} to HTML."
--- a/lib/gollum-lib/filter/render.rb
+++ b/lib/gollum-lib/filter/render.rb
@@ -6,7 +6,7 @@ class Gollum::Filter::Render < Gollum::Filter
       working_dir = Pathname.new(@markup.wiki.path).join(@markup.dir)
       working_dir = working_dir.exist? ? working_dir.to_s : '.'
       Dir.chdir(working_dir) do
-        data = GitHub::Markup.render(@markup.name, data)
+        data = GitHub::Markup.render_s(Gollum::Markup.formats[@markup.format][:name].downcase.to_sym, data)
       end
       if data.nil?
         raise "There was an error converting #{@markup.name} to HTML."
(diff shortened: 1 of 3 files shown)


dmyates:
This doesn't work at all (and I believe I have been misunderstanding the relationship between `Gollum::Markup.formats` and `Github::Markup` this whole time). Back to the drawing board.

dmyates:
Alright, this should be better. The build is still failing, but `bundle exec rake test` succeeds when I run it locally.

dometto:
Thanks @dmyates! </verylate>

The remaining test failure was due to a timeout. Restarted the build, confident that it will pass now.

Have you checked that the symbols we get from `@markup.format` correspond to the relevant symbols in `github-markup`, for each supported markup?

dmyates:
```
module Gollum
  class Markup
    # <SNIP>
    register(:markdown,  "Markdown", :regexp => /md|mkdn?|mdown|markdown/)
    register(:textile,   "Textile")
    register(:rdoc,      "RDoc")
    register(:org,       "Org-mode")
    register(:creole,    "Creole", :reverse_links => true)
    register(:rest,      "reStructuredText", :regexp => /re?st(\.txt)?/)
    register(:asciidoc,  "AsciiDoc")
    register(:mediawiki, "MediaWiki", :regexp => /(media)?wiki/, :reverse_links => true)
    register(:pod,       "Pod")
    register(:txt,       "Plain Text")
  end
end
```

```
module GitHub
  module Markups
    # all of supported markups:
    MARKUP_ASCIIDOC = :asciidoc
    MARKUP_CREOLE = :creole
    MARKUP_MARKDOWN = :markdown
    MARKUP_MEDIAWIKI = :mediawiki
    MARKUP_ORG = :org
    MARKUP_POD = :pod
    MARKUP_RDOC = :rdoc
    MARKUP_RST = :rst
    MARKUP_TEXTILE = :textile
    MARKUP_POD6 = :pod6
  end
```

Looks like I missed a couple. `github-markup` uses `:rst`, but Gollum uses `:rest`, and `github-markup` has no plain text option.

dometto:
The plain text use case should be ok, since `.txt` files aren't being rendered at all (and therefore not passed to `github-markup`). The `:rest` symbol could be changed to `:rst`, but that would require two small changes to `gollum`, as well:
* [here](https://github.com/gollum/gollum/blob/5.x/lib/gollum/views/page.rb#L212)
* [here](https://github.com/gollum/gollum/blob/5.x/lib/gollum/public/gollum/javascript/editor/gollum.editor.js.erb#L28).

If you could open a PR for that against the gollum `5.x` branch, I'll merge this first, and then the new PR.

Additionally, it would be great if you could add the following test to this PR (for example in `test_markup.rb`):

```ruby
  test 'github-markup knows about gollum markups' do
    markups_with_render_filter = Gollum::Markup.formats.select do |k, v|
      case v[:skip_filters]
      when Array
        !v[:skip_filters].include?(:Render)
      when Proc
        !v[:skip_filters].call(:Render)
      else
        true
      end
    end
    markups_with_render_filter.each do |name, info|
      assert ::GitHub::Markup.markups.key?(name), "GitHub::Markup does not know about format #{name}"
    end
  end
```

dometto:
Thanks!

PR #333, opened 2019-08-18
Allow setting of subpages on PreviewPage

dometto:
Define `#set_sidebar` etc. on `PreviewPage` class

### DIFF
--- a/test/test_wiki.rb
+++ b/test/test_wiki.rb
@@ -125,6 +125,10 @@ context "Wiki page previewing" do
     assert_equal "Test.md", page.filename
     assert_equal "Test", page.name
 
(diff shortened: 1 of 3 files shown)


test/test_wiki.rb
```diff
@@ -125,6 +125,10 @@
     assert_equal "Test.md", page.filename
     assert_equal "Test", page.name
 
+    # Getting and setting subpage contents
+    assert page.set_sidebar("*Waa*\n")
```
bartkamphorst: Could we make `*Waa*` into a LOTR reference or something instead? 😇

bartkamphorst:
Nice solution. 👍

PR #334, opened 2019-08-19
Actually respect pagination options

dometto:
Pagination methods were not respecting custom `:skip` passed to `log_pagination_options`


bartkamphorst:
Sweet. 👍

PR #339, opened 2019-09-10
Make code filter conform to GitHub

dometto:
According to the [GFM spec](https://github.github.com/gfm/#fenced-code-blocks), fenced code blocks can't have more than three spaces indent, and opening and closing tags don't have to have the same amount of indent.

Closing tags must also have at least as many characters as opening tags. However, gollum only supports exactly three backticks for fenced codeblocks on all pages. This PR implements the minimum length for end tags for tilde-style `~~~` fenced code blocks on markdown, however.

Also solved a bug that was causing the info-string (codeblock language) not to be correctly parsed for tilde fenced blocks: `~~~ruby` was illegal, `~~~.ruby` expected, but this is not correct according to the current GFM spec.

Question:

> However, gollum only supports exactly three backticks for fenced codeblocks on all pages.

Does that strike us as sane behavior, or do we want fenced codeblocks that can use any number of backticks (higher than three, and at least as long end tag as star tag)?

### DIFF
--- a/lib/gollum-lib/filter/code.rb
+++ b/lib/gollum-lib/filter/code.rb
@@ -20,25 +20,21 @@ class Gollum::Filter::Code < Gollum::Filter
         "#{Regexp.last_match[3]}#{cache_codeblock(Regexp.last_match[4], Regexp.last_match[5])}"
       end
     when :markdown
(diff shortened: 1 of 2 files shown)


bartkamphorst:
> Does that strike us as sane behavior?

Yes, for me the syntax is much clearer with exactly three backticks instead of any number over two.

dometto:
👍 was hoping you'd say that!

PR #343, opened 2019-09-18
Advanced search. Resolves #166

### LINKED ISSUE #166: Better Search

dometto:
Requires https://github.com/gollum/rugged_adapter/pull/37

### DIFF
--- a/lib/gollum-lib/wiki.rb
+++ b/lib/gollum-lib/wiki.rb
@@ -512,7 +512,7 @@ module Gollum
       results = @repo.git.grep(search_terms, options) do |name, data|
         result = {:count => 0}
         result[:name] = extract_page_file_dir(name)
-        result[:count] += 1 if result[:name] =~ /#{search_terms_regex}/i
+        result[:filename_count] = result[:name].scan(/#{search_terms_regex}/i).size
         found_in_blob = []
         if data
           data.scan(query) do |match|
@@ -521,7 +521,7 @@ module Gollum
           end
         end
         result[:context] = found_in_blob
-        result[:count] == 0 ? nil : result
+        ((result[:count] + result[:filename_count]) == 0) ? nil : result
       end
       [results, return_terms ? search_terms : nil]
     end
--- a/lib/gollum-lib/wiki.rb
+++ b/lib/gollum-lib/wiki.rb
@@ -511,7 +511,7 @@ module Gollum
         result[:name] = extract_page_file_dir(result[:name])
         result
       end
-      return_terms ? [results, search_terms] : results
+      [results, return_terms ? search_terms : nil]
     end
 
     # Public: All of the versions that have touched the Page.
--- a/lib/gollum-lib/wiki.rb
+++ b/lib/gollum-lib/wiki.rb
@@ -504,7 +504,7 @@ module Gollum
     # query - The string to search for
     #
     # Returns an Array with Objects of page name and count of matches
-    def search(query, return_terms = false)
+    def search(query)
       options = {:path => page_file_dir, :ref => ref}
       search_terms = query.scan(/"([^"]+)"|(\S+)/).flatten.compact.map {|term| Regexp.escape(term)}
       search_terms_regex = search_terms.join('|')
@@ -513,17 +513,16 @@ module Gollum
         result = {:count => 0}
         result[:name] = extract_page_file_dir(name)
         result[:filename_count] = result[:name].scan(/#{search_terms_regex}/i).size
-        found_in_blob = []
+        result[:context] = []
         if data
           data.scan(query) do |match|
-            found_in_blob << match.first
+            result[:context] << match.first
             result[:count] += match.first.scan(/#{search_terms_regex}/i).size
           end
         end
-        result[:context] = found_in_blob
         ((result[:count] + result[:filename_count]) == 0) ? nil : result
       end
-      [results, return_terms ? search_terms : nil]
+      [results, search_terms]
     end
 
     # Public: All of the versions that have touched the Page.
--- a/lib/gollum-lib/wiki.rb
+++ b/lib/gollum-lib/wiki.rb
@@ -504,31 +504,14 @@ module Gollum
     # query - The string to search for
     #
     # Returns an Array with Objects of page name and count of matches
-    def search(query)
+    def search(query, return_terms = false)
       options = {:path => page_file_dir, :ref => ref}
-      results = {}
-      @repo.git.grep(query, options).each do |hit|
-        name = hit[:name]
-        count = hit[:count]
-        # Remove ext only from known extensions.
-        # test.pdf => test.pdf, test.md => test
-        file_name = Page::valid_page_name?(name) ? name.chomp(::File.extname(name)) : name
-        results[file_name] = count.to_i
+      search_terms = query.scan(/"([^"]+)"|(\S+)/).flatten.compact.map {|term| Regexp.escape(term)}
+      results = @repo.git.grep(search_terms, options).map do |result|
+        result[:name] = extract_page_file_dir(result[:name])
+        result
       end
-
-      # Use ls_files '*query*' to search for file names. Grep only searches file content.
-      # Spaces are converted to dashes when saving pages to disk.
-      @repo.git.ls_files(query, options).each do |path|
-        # Remove ext only from known extensions.
-        file_name          = Page::valid_page_name?(path) ? path.chomp(::File.extname(path)) : path
-        # If there's not already a result for file_name then
-        # the value is nil and nil.to_i is 0.
-        results[file_name] = results[file_name].to_i + 1;
-      end
-
-      results.map do |key, val|
-        { :count => val, :name => key }
-      end
+      return_terms ? [results, search_terms] : results
     end
 
     # Public: All of the versions that have touched the Page.
@@ -761,6 +744,10 @@ module Gollum
       else
         result[0] == '/' ? result[1..-1] : result
       end
+    end
+
+    def extract_page_file_dir(path)
+      @page_file_dir ? path[@page_file_dir.length+1..-1] : path
     end
 
     def write(path, data, commit = {})
--- a/lib/gollum-lib/wiki.rb
+++ b/lib/gollum-lib/wiki.rb
@@ -507,9 +507,21 @@ module Gollum
     def search(query, return_terms = false)
       options = {:path => page_file_dir, :ref => ref}
       search_terms = query.scan(/"([^"]+)"|(\S+)/).flatten.compact.map {|term| Regexp.escape(term)}
-      results = @repo.git.grep(search_terms, options).map do |result|
-        result[:name] = extract_page_file_dir(result[:name])
-        result
+      search_terms_regex = search_terms.join('|')
+      query = /^(.*(?:#{search_terms_regex}).*)$/i
+      results = @repo.git.grep(search_terms, options) do |name, data|
+        result = {:count => 0}
+        result[:name] = extract_page_file_dir(name)
+        result[:count] += 1 if result[:name] =~ /#{search_terms_regex}/i
+        found_in_blob = []
+        if data
+          data.scan(query) do |match|
+            found_in_blob << match.first
+            result[:count] += match.first.scan(/#{search_terms_regex}/i).size
+          end
+        end
+        result[:context] = found_in_blob
+        result[:count] == 0 ? nil : result
       end
       [results, return_terms ? search_terms : nil]
     end
--- a/test/test_search.rb
+++ b/test/test_search.rb
@@ -8,13 +8,14 @@ context 'Wiki search' do
 Hobbits are small. Hobbits live in Hobbiton.
 One thing Hobbits do not like is Orcs.
 Something a hobbit does like is pastry.
+Some other words.
 EOF
 
   SEARCH_TEST_LINES = SEARCH_TEST_PAGE.split(/[\n\r]+/)
 
   setup do
     @path = cloned_testpath('examples/lotr.git')
-    @wiki = Gollum::Wiki.new(@path, :page_file_dir => 'Mordor')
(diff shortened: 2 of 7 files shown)


dometto:
Hope the specs make clear what the new behavior is.

test/test_search.rb
```diff
@@ -0,0 +1,96 @@
...
+    context = results.first[:context]
+    assert_not_nil context
+    lines = SEARCH_TEST_PAGE.split("\n")
+    assert context.include?(SEARCH_TEST_LINES[0])
+    assert context.include?(SEARCH_TEST_LINES[1])
+    assert context.include?(SEARCH_TEST_LINES[2])
+    assert !context.include?(SEARCH_TEST_LINES[3])
+  end
+
+  test 'search returns escaped search terms on demand' do
```
dometto: I thought this behavior might be nice so we can potentially pass the list of search terms on to the Javascript, and use that to highlight the searched for terms in the search result screen.

lib/gollum-lib/wiki.rb
```diff
@@ -504,31 +504,14 @@ def size(ref = nil)
...
-      end
-
-      results.map do |key, val|
-        { :count => val, :name => key }
+      search_terms = query.scan(/"([^"]+)"|(\S+)/).flatten.compact.map {|term| Regexp.escape(term)}
+      results = @repo.git.grep(search_terms, options).map do |result|
+        result[:name] = extract_page_file_dir(result[:name])
+        result
       end
+      return_terms ? [results, search_terms] : results
```
bartkamphorst: I would prefer it if the returned object is always of the same data structure, so I would have the else clause return `[results, nil]` or `[results, []]` and always assume that search returns a two-placed array.

lib/gollum-lib/wiki.rb
```diff
@@ -504,31 +504,14 @@ def size(ref = nil)
...
-      @repo.git.grep(query, options).each do |hit|
-        name = hit[:name]
-        count = hit[:count]
-        # Remove ext only from known extensions.
-        # test.pdf => test.pdf, test.md => test
-        file_name = Page::valid_page_name?(name) ? name.chomp(::File.extname(name)) : name
-        results[file_name] = count.to_i
-      end
-
-      # Use ls_files '*query*' to search for file names. Grep only searches file content.
```
bartkamphorst: Is this reasoning no longer valid? Using `ls_files` to search for file names?

test/test_search.rb
```diff
@@ -0,0 +1,96 @@
+# ~*~ encoding: utf-8 ~*~
+path = File.join(File.dirname(__FILE__), 'helper')
+require File.expand_path(path)
+
+context 'Wiki search' do
+
+  SEARCH_TEST_PAGE = <<EOF
+Hobbits are small. Hobbits live in Hobbiton.
```
bartkamphorst: Not all hobbits live in Hobbiton, but Bilbo does.

test/test_search.rb
```diff
@@ -0,0 +1,96 @@
...
+One thing Hobbits do not like is Orcs.
+Something a hobbit does like is pastry.
+Some other words.
```
bartkamphorst: How about some utf-8 instead? ` ᚠᛇᚻ᛫ᛒᛦᚦ᛫ᚠᚱᚩᚠᚢᚱ᛫ᚠᛁᚱᚪ᛫ᚷᛖᚻᚹᛦᛚᚳᚢᛗ`

test/test_search.rb
```diff
@@ -0,0 +1,96 @@
...
+  test 'search results should make the content/filename search additive' do
+    # There is a file that contains the word 'foo' and is called 'foo', so it should
+    # have a count of 2, not 1...
+    results = @wiki.search('foo')
+    assert_equal 2, results.first[:count]
+  end
+
+  test 'search results should not include files that do not match the query' do
+    results = @wiki.search('foo')
+    assert_equal 1, results.size
```
bartkamphorst: I guess you should also explicitly assert that a certain file is not present, given the name of this test.

test/test_search.rb
```diff
@@ -0,0 +1,96 @@
...
+
+  test 'search should respect quoted search terms' do
+    results = @wiki.search('like is pastry')
+    hobbit_info_result = results.find {|result| result[:name] == 'Hobbit Info.md'}
+    assert_equal 5, hobbit_info_result[:count]
+    results = @wiki.search('"like is" pastry')
+    assert_equal 3, results.first[:count]
+  end
+
+  test 'search should be case insensitive' do
```
bartkamphorst: The implementation of this test doesn't make case insensitivity intuitively clear. I would expect something like searching for downcase and finding uppercase.

bartkamphorst: Note that this comment does not concern the actual workings of the test but just how the test "reads".

test/test_search.rb
```diff
@@ -0,0 +1,96 @@
...
+    results = @wiki.search('Hobbit')
+    hobbit_info_result = results.find {|result| result[:name] == 'Hobbit Info.md'}
+    assert_equal 6, hobbit_info_result[:count]
+  end
+
+  test 'search should return context for hits' do
```
bartkamphorst: How much context is provided? And how is the amount of context determined?

test/test_search.rb
bartkamphorst: Agreed.

test/test_search.rb
```diff
@@ -0,0 +1,96 @@
...
+    assert_not_nil context
+    lines = SEARCH_TEST_PAGE.split("\n")
+    assert context.include?(SEARCH_TEST_LINES[0])
+    assert context.include?(SEARCH_TEST_LINES[1])
+    assert context.include?(SEARCH_TEST_LINES[2])
+    assert !context.include?(SEARCH_TEST_LINES[3])
+  end
+
+  test 'search returns escaped search terms on demand' do
+    result, search_terms = @wiki.search('foo|bar', true)
```
bartkamphorst: Are exclusive disjunctions (and conjunctions for that matter) respected?

test/test_search.rb
```diff
@@ -0,0 +1,96 @@
...
+    assert_equal 'foo\|bar', search_terms.first
+  end
+
+  test 'search should respect page_file_dir' do
```
bartkamphorst: Great to have this test!

bartkamphorst: Perhaps also explicitly check that 'Rivendell' is not in the returned file path?

test/test_search.rb
dometto: I'm afraid I can't correct that without breaking the tests 😄

test/test_search.rb
dometto: So geeky! I like it.

test/test_search.rb
dometto: The amount of context is determined in the adapter. The way I've implemented grep in the rugged adapter now just returns the line on which a hit was found as context. I could add an assertion that context should be exactly one line if we want to make that the required behavior.

test/test_search.rb
dometto: This test should maybe be a bit clearer: what it does is *not* respect the `|` by escaping it. :) That is: we shouldn't let the user attempt to construct a regexp, because that will break our own scanning regex in the adapter.

I think XAND is supported via the `"quote words"` functionality: see the "respect quoted search terms" test. Or were you thinking of something else?

Searching disjunctions in this implementation is limited to this: the search string `foo bar XYZ` will score a hit for each line that contains either `foo`, `bar` or `XYZ`.

If you want to search for files that contain *both* `foo`, and (`bar` XOR `XYZ`), you *could* use the following search string:  `"foo bar" "foo XYZ" "XYZ foo" "bar foo"`

What you can't do is search for files that contain *both* `foo`, and (`bar` XOR `XYZ`). That would be cool, but would make the implementation much more complicated I think (not saying we shouldn't do it -- what do you think?)

lib/gollum-lib/wiki.rb
dometto: No, when looping through the repo the adapter already has the filenames, so this additional loop through the repo is unnecessary and expensive. Ergo we'll be checking whether the filename is a hit in the adapter (I'll update the adapter spec suite to reflect this -- in fact I think most tests that are now included here should move to the adapter spec suite).

test/test_search.rb
dometto: @bartkamphorst what do you think about this? Is the current implementation intuitive, and should we implement more complex search patterns?

bartkamphorst: First of all, I think the new behavior will be an improvement over the previous behavior. My question pertained mostly to your second example of `foo AND (bar XOR XYZ)`, or even more generally to `(bar XOR foo)` and to `bar AND NOT foo`. The latter case in particular would be a great addition, so that you would be able to search for "Hobbit AND NOT Bilbo". 😉 I don't know how feasible any of those options are though.

dometto: I agree they would be cool, but constructing regexes based on user input like that feels like re-inventing the wheel a bit. It would be great if there were a generic solution for generating regexes out of user search strings like that -- or alternatively, if we could just let the user type regexes.

bartkamphorst: I've been looking but I haven't found anything out of the box that does what we want without too much overhead. Here's [one approach](https://www.sitepoint.com/community/t/need-help-converting-user-friendly-search-string-to-regex/85747) though that might serve as inspiration.

bartkamphorst: > or alternatively, if we could just let the user type regexes.

There's some security issues with this approach, for example DOS attacks if the expression takes long to resolve.

I'm surprised there's no standard way of dealing with this other than using full-blown search libraries.

dometto: See https://gist.github.com/dometto/c9ffadfaacd8f50c31d69412a65ba64f for most advanced implementation (supporting NOT and AND) I can get to work without implementing a full-fledged parser. Note that I couldn't get disjunctions within search terms to work :(

test/test_search.rb
bartkamphorst: Yes, it would be good to have that assertion I think.

test/test_search.rb
bartkamphorst: I have changed my mind about this. I think we should just start with the scope of the current PR, with context, multiple keyword search and exact matches. 👍

dometto:
Removed the matching logic from the adapters and implemented it here. Also refactored based on @bartkamphorst's earlier suggested.

lib/gollum-lib/wiki.rb
```diff
@@ -504,31 +504,26 @@ def size(ref = nil)
...
+      search_terms_regex = search_terms.join('|')
+      query = /^(.*(?:#{search_terms_regex}).*)$/i
+      results = @repo.git.grep(search_terms, options) do |name, data|
+        result = {:count => 0}
+        result[:name] = extract_page_file_dir(name)
+        result[:filename_count] = result[:name].scan(/#{search_terms_regex}/i).size
+        found_in_blob = []
+        if data
+          data.scan(query) do |match|
+            found_in_blob << match.first
```
bartkamphorst: The variable name `found_in_blob` is a little confusing, especially since it also doubles as context. Is there a more expressive name you could give it?

lib/gollum-lib/wiki.rb
```diff
@@ -504,31 +504,26 @@ def size(ref = nil)
...
-        # the value is nil and nil.to_i is 0.
-        results[file_name] = results[file_name].to_i + 1;
-      end
-
-      results.map do |key, val|
-        { :count => val, :name => key }
+      search_terms = query.scan(/"([^"]+)"|(\S+)/).flatten.compact.map {|term| Regexp.escape(term)}
+      search_terms_regex = search_terms.join('|')
+      query = /^(.*(?:#{search_terms_regex}).*)$/i
+      results = @repo.git.grep(search_terms, options) do |name, data|
```
bartkamphorst: I like this a lot!

lib/gollum-lib/wiki.rb
```diff
@@ -504,31 +504,26 @@ def size(ref = nil)
...
+        result[:filename_count] = result[:name].scan(/#{search_terms_regex}/i).size
+        found_in_blob = []
+        if data
+          data.scan(query) do |match|
+            found_in_blob << match.first
+            result[:count] += match.first.scan(/#{search_terms_regex}/i).size
+          end
+        end
+        result[:context] = found_in_blob
+        ((result[:count] + result[:filename_count]) == 0) ? nil : result
```
bartkamphorst: What is this line supposed to do? It is not a return statement but it's also not setting any variable it seems.

lib/gollum-lib/wiki.rb
```diff
@@ -504,31 +504,26 @@ def size(ref = nil)
...
       end
+      [results, return_terms ? search_terms : nil]
```
bartkamphorst: 👍

lib/gollum-lib/wiki.rb
dometto: It is supposed to function as a return statement, but not sure if you can use explicit `return` in a block. Either way the idea is to only store results for files that actually contain a hit (either in-text or filename).

lib/gollum-lib/wiki.rb
dometto: Actually, if we're always returning a two-place array anyway, why not just always return `search_terms`, as well? :)

lib/gollum-lib/wiki.rb
dometto: Changed to `hits`

dometto:
Had to require the dev versions of the adapters in the Gemfile and set the default adapter to rugged to make the tests pass.

PR #345, opened 2019-09-26
Add Note() and Warn() flash macros. Add Octicon() macro. Improve Macro error rendering.

bartkamphorst:
This PR adds two macros for adding notes or warnings to a wiki page. The syntax for these is straightforward, each takes a String (see the example below). In addition, this PR adds a macro for adding any of the [Octicons](https://octicons.github.com/) to a wiki page. This macro takes a String argument for determining which octicon to render, plus two optional arguments for setting the height and width of the image. 

Screen Shot 2019-09-26 at 20 14 59

Finally, the PR also enhances the way Macro errors are rendered on a page. The rendered result of the above is as follows.

Screen Shot 2019-09-26 at 09 58 36

### DIFF
--- a/lib/gollum-lib/filter/macro.rb
+++ b/lib/gollum-lib/filter/macro.rb
@@ -1,4 +1,5 @@
 # ~*~ encoding: utf-8 ~*~
+require 'octicons'
 
 # Replace specified tokens with dynamically generated content.
 class Gollum::Filter::Macro < Gollum::Filter
@@ -47,7 +48,9 @@ class Gollum::Filter::Macro < Gollum::Filter
         begin
           Gollum::Macro.instance(macro, @markup.wiki, @markup.page).render(*args)
         rescue StandardError => e
-          "!!!Macro Error: #{e.message}!!!"
+          icon = Octicons::Octicon.new('zap', {width: 24, height: 24})
+          icon.options[:class] << ' mr-2'
+          "<div class='flash flash-error'>#{icon.to_svg}!!!Macro Error for #{macro}: #{e.message}!!!</div>"
         end
       end
     end
--- a/lib/gollum-lib/filter/macro.rb
+++ b/lib/gollum-lib/filter/macro.rb
@@ -50,7 +50,7 @@ class Gollum::Filter::Macro < Gollum::Filter
         rescue StandardError => e
           icon = Octicons::Octicon.new('zap', {width: 24, height: 24})
           icon.options[:class] << ' mr-2'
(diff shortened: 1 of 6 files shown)


bartkamphorst:
The remaining test failures are unrelated to this PR (they have to do with search).

dometto:
Nice addition! Looks good to me. One minor question in the comments.

lib/gollum-lib/filter/macro.rb
```diff
@@ -47,7 +48,9 @@ def process(data)
         begin
           Gollum::Macro.instance(macro, @markup.wiki, @markup.page).render(*args)
         rescue StandardError => e
-          "!!!Macro Error: #{e.message}!!!"
+          icon = Octicons::Octicon.new('zap', {width: 24, height: 24})
+          icon.options[:class] << ' mr-2'
+          "<div class='flash flash-error'>#{icon.to_svg}!!!Macro Error for #{macro}: #{e.message}!!!</div>"
```
dometto: Now that there's a nice shiny red alert box, couldn't we get rid of the triple exclamation marks on both sides?

bartkamphorst: Yup, will do.

dometto:
Also, could you document these new additions in the wiki? (Either in the 5.x Release Notes or a separate Macros page) Thanks!

bartkamphorst:
> Looks good to me. 

Thanks. The only thing I'm thinking of changing now is to allow users of the macro to set a different octicon in the flash box, or to leave the octicon out. So the syntax would then be:
```ruby
<<Note("Notice the different octicon", "bell")>>
<<Note("Or notice this box without an octicon", nil)>>
```
The second parameter would be optional so the default use would still be very straightforward with passing a String. Thoughts @dometto ?

bartkamphorst:
> could you document these new additions in the wiki?

Will do once the syntax is set and the changes have been merged.

dometto:
> The second parameter would be optional so the default use would still be very straightforward with passing a String.

Souds cool to me, but if you're going to make the Macros more complex, maybe you should consider doing some subclassing to avoid code duplication?

bartkamphorst:
> Souds cool to me

Here's what the three different options would look like.

Screen Shot 2019-09-27 at 13 44 52

> maybe you should consider doing some subclassing to avoid code duplication?

The alternative is to keep `Warn()` as is, but only change Note(). But even if we give `Warn()` the same options, I'm not too worried about code duplication yet for these simple macros.

dometto:
> The alternative is to keep Warn() as is, but only change Note().

That also seems plausible to me!

bartkamphorst:
@dometto See the latest commits.

PR #347, opened 2019-09-30
Fix image with frame alignment

bartkamphorst:
This PR utilizes `Primer.css` classes to address https://github.com/gollum/gollum/issues/838.  The tests are still failing due to changes in the class names, but I would like input regarding the expected functionality before I fix those class names in the tests. The current functionality in this PR is as follows:
* Adding `float` will create a floating image, with text wrapping around the image. Floats will be aligned left by default, unless `align=right` is specified. Left and right are the only two valid options for floating images.
* Adding `frame` will add a border to the image. If `alt=alt text` is set, the image will be accompanied by a caption below the image, with text alignment set to the alignment of the image (left, center, or right).
* Images with and without frames can be aligned left, center, or right (e.g.,`align=center`). If alignment is set to left, center, or right, the image will be "containered".
* Images without `float` or alignment `{left, center, right}` will not be "containered". For these images, the `img` attribute `align` may be set to one of `top texttop middle absmiddle bottom absbottom baseline`. I'm not sure what this does or is supposed to do, but that followed from the previous implementation.


bartkamphorst:
@penguinpowernz Would appreciate your input on this PR as well!

dometto:
Nice! That was quite a puzzle with the new primer classes.

> If alignment is set to left, center, or right, the image will be "containered".

Not sure what the reason for that is, but as it appears to be the current behavior, I am fine with it. 👍

dometto:
All the other behavior makes sense to me!

bartkamphorst:
Thanks. Would you mind giving it a spin locally to see if everything works as you would expect?

dometto:
`align left|center|right` seems to be working, as is text-wrap with `float`. But `frame` and `alt` only seem to work when `align` is also set, and not on their own:

Screenshot 2019-09-30 at 14 44 39

Was that already the case in the previous implementation? Would it be difficult to change?

bartkamphorst:
> But frame and alt only seem to work when align is also set, and not on their own

Fix in https://github.com/gollum/gollum-lib/pull/347/commits/2b0712d4efbef9a67fff7e591cef3480f8b29c5d. It is now possible again to give an image a frame / border without setting `align` explicitly (it defaults to left alignment).

dometto:
Sweet!

PR #348, opened 2019-10-01
Fix Navigation Macro with page_file_dir

dometto:
See https://github.com/gollum/gollum/issues/1428


bartkamphorst:
`url_path` for the win.

PR #350, opened 2019-10-05
Add support for custom redirects (e.g., after renames). 

bartkamphorst:
Relates to https://github.com/gollum/gollum/issues/1023. This PR should be considered together with [this gollum PR](https://github.com/gollum/gollum/pull/1433).

It works by having a protected `.redirects.gollum` file that contains a YAML hash with redirects. For example: 
```yaml
---
Home.old.md: Home.md
subdir/Home.md: Home.md
tobemoved.md: wasmoved.md
```

This PR also 
* adds a public method `#overwrite_file` to wiki objects. 
* removes protected files from file/page listings.


dometto:
Very nice PR!

lib/gollum-lib/redirects.rb
```diff
@@ -0,0 +1,28 @@
...
+  class Redirects
+        
+    def self.load(wiki)
+      file = wiki.file(REDIRECTS_FILE)
+      redirects = {}
+      if file
+        begin
+          redirects = YAML.load(file.raw_data)
+        rescue YAML::Error
+          # TODO handle error
```
dometto: Method now returns `nil` when an exception is rescued. Will that go down well in gollum?

lib/gollum-lib/wiki.rb
```diff
@@ -240,6 +240,26 @@ def write_page(name, format, data, commit = {})
...
+    #          :parent    - Optional Gollum::Git::Commit parent to this update.
+    #          :tree      - Optional String SHA of the tree to create the
+    #                       index from.
+    #          :committer - Optional Gollum::Committer instance.  If provided,
+    #                       assume that this operation is part of batch of
+    #                       updates and the commit happens later.
+    # Returns the String SHA1 of the newly written version, or the
+    # Gollum::Committer instance if this is part of a batch update
+    def overwrite_file(name, data, commit = {})
+      write(merge_path_elements(nil, name, nil), data, commit, force_overwrite = true)
```
dometto: 👍

lib/gollum-lib/wiki.rb
```diff
@@ -579,6 +599,16 @@ def history_sanitizer
       @history_sanitizer ||= history_sanitization.to_sanitize
     end
 
+    def redirects
+      ::Gollum::Redirects.load(self)
+    end
+    
+    def add_redirect(old_path, new_path)
+      redirects = ::Gollum::Redirects.load(self)
```
dometto: with the above `redirects` method, isn't this line superfluous?

lib/gollum-lib/wiki.rb
```diff
@@ -579,6 +599,16 @@ def history_sanitizer
       @history_sanitizer ||= history_sanitization.to_sanitize
     end
 
+    def redirects
+      ::Gollum::Redirects.load(self)
```
dometto: I'm not against it per se, but find it a bit surprising that `Gollum::Redirects` is a class that doesn't have instances. Is it conceptually right to think of Redirects as a class that takes a reference to a wiki, rather than a Wiki having a redirects property? Also, for instance, if we'd want to build in caching in the future (i.e. not reading the YAML store from the repo unless `HEAD` has changed)? Again, I don't feel strongly.

lib/gollum-lib/redirects.rb
bartkamphorst: It shouldn't, since `redirects` is returned and `redirects` is initialized as `{}`.

lib/gollum-lib/wiki.rb
bartkamphorst: Fair points. In response, I've re-implemented `::Gollum::Redirects` as a Hash extension Module that adds the following methods to the redirects hash: `init(wiki), stale?, load, dump`. I also followed your suggestion to only reload `@redirects` if the repo `HEAD` changed since initialization of `@redirects`. Have a look and tell me what you think!

lib/gollum-lib/wiki.rb
bartkamphorst: Fixed in new implementation.

lib/gollum-lib/wiki.rb
```diff
@@ -579,6 +599,25 @@ def history_sanitizer
       @history_sanitizer ||= history_sanitization.to_sanitize
     end
 
+    def redirects
+      if @redirects.nil? || @redirects.stale?
+        @redirects = {}.extend(::Gollum::Redirects)
```
dometto: Nice!

bartkamphorst: Thanks. ☺️

bartkamphorst:
> Maybe add one more test for the Redirects.stale? method, that also checks whether the redirects file is reloaded when stale? is true?

Something like this?

```ruby
  test "#redirects reloads the redirects hash when the cache has become stale" do
    @wiki.add_redirect('oldpage.md', 'newpage.md')
    redirects = @wiki.redirects
    object_id1 = redirects.object_id
    assert_equal 'newpage.md', redirects['oldpage.md']
    # Test that the cache works by calling #redirects again
    assert_equal object_id1, @wiki.redirects.object_id
    # Overwriting the redirects file changes HEAD, turning the redirects cache stale
    @wiki.overwrite_file('.redirects.gollum', {'Home.old.md' => 'Home.md'}.to_yaml)
    redirects = @wiki.redirects
    object_id2 = redirects.object_id
    assert_not_equal object_id1, object_id2
    assert_equal nil, redirects['oldpage.md']
    assert_equal 'Home.md', redirects['Home.old.md']
  end
```

dometto:
Sure!

bartkamphorst:
Done.

PR #351, opened 2019-10-10
Replace Sanitize with Loofah. Resolves #320

### LINKED ISSUE #320: Depends on an old version of sanitize
couldnt we at least allow a newer version for the non java build?

dometto:
Sanitize is no longer being maintained for JRuby (see #320). [Loofah](https://github.com/flavorjones/loofah) is the sanitizer used in the rails asset pipeline. It wasn't all that difficult to plug it in and get the tests passing, and I think it's actually a bit easier to understand what's happening than with Sanitize.

Loofah has some sane defaults for safe HTML5 elements: https://github.com/flavorjones/loofah/blob/master/lib/loofah/html5/safelist.rb#L47, so we don't have to maintain that ourselves. End users can still override these settings by overriding `Loofah::HTML5::SafeList`.

### DIFF
--- a/lib/gollum-lib/filter/tags.rb
+++ b/lib/gollum-lib/filter/tags.rb
@@ -173,14 +173,8 @@ class Gollum::Filter::Tags < Gollum::Filter
   # Return the String HTML if the tag is a valid external link tag or
   # nil if it is not.
   def process_external_link_tag(url, pretty_name = nil)
-    accepted_protocols = @markup.wiki.sanitization.protocols['a']['href'].dup
-    if accepted_protocols.include?(:relative)
-      accepted_protocols.select!{|protocol| protocol != :relative}
-      regexp = %r{^((#{accepted_protocols.join("|")}):)?(//)}
-    else
-      regexp = %r{^((#{accepted_protocols.join("|")}):)}
-    end
-    if url =~ regexp
+    @accepted_protocols_regex ||= %r{^((#{::Gollum::Sanitization.accepted_protocols.join('|')}):)?(//)} 
+    if url =~ @accepted_protocols_regex
       generate_link(url, pretty_name, nil, :external)
     else
       nil
--- a/lib/gollum-lib/page.rb
+++ b/lib/gollum-lib/page.rb
@@ -101,7 +101,7 @@ module Gollum
     #
     # Returns the fully sanitized String title.
     def title
-      Sanitize.clean(name).strip
+      @wiki.sanitizer.clean(name).strip
     end
 
     # Public: Determines if this is a sub-page
--- a/lib/gollum-lib/sanitization.rb
+++ b/lib/gollum-lib/sanitization.rb
@@ -1,378 +1,65 @@
-# ~*~ encoding: utf-8 ~*~
+::Loofah::HTML5::SafeList::ACCEPTABLE_PROTOCOLS.add('apt')
+
 module Gollum
-  # Encapsulate sanitization options.
-  #
-  # This class does not yet support all options of Sanitize library.
-  # See http://github.com/rgrove/sanitize/.
   class Sanitization
 
-    # Default whitelisted elements required for MathML. See https://developer.mozilla.org/en-US/docs/Web/MathML/Element
-    # For some help on generating MATHML_ELEMENTS and MATHML_ATTRS, see https://gist.github.com/dometto/52d9cb8b45d68bfc7665e5e6683b75a0
-    MATHML_ELEMENTS = [
-      'math', 'maction', 'maligngroup', 'malignmark', 'menclose',
-      'merror', 'mfenced', 'mfrac', 'mglyph', 'mi', 'mlabeledtr',
-      'mlongdiv', 'mmultiscripts', 'mn', 'mo', 'mover', 'mpadded',
-      'mphantom', 'mroot', 'mrow', 'ms', 'mscarries', 'mscarry',
-      'msgroup', 'msline', 'mspace', 'msqrt', 'msrow', 'mstack',
-      'mstyle', 'msub', 'msup', 'msubsup', 'mtable', 'mtd',
-      'mtext', 'mtr', 'munder', 'munderover', 'semantics'
-    ].freeze
+    @@accepted_protocols = ::Loofah::HTML5::SafeList::ACCEPTABLE_PROTOCOLS.to_a.freeze
 
-
-    # Default whitelisted attributes required for MathML. See https://developer.mozilla.org/en-US/docs/Web/MathML/Attribute 
-    MATHML_ATTRS = {
-     'math'=>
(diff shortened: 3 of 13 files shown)


dometto:
See here for a comparison: https://github.com/rgrove/sanitize/blob/master/COMPARISON.md
  * The benchmarks there compare Sanitize vs an older Loofah (Sanitize wins)
  * @bartkamphorst  more importantly, what do you think of the other differences? 
And here for more recent benchmarks (: https://gist.github.com/flavorjones/170193
  * Benchmarks: more recent Loofah wins against Sanitize

dometto:
Loofah is tested on JRuby, but see [here](https://github.com/flavorjones/loofah/issues/88)

lib/gollum-lib/sanitization.rb
```diff
@@ -1,378 +1,52 @@
-# ~*~ encoding: utf-8 ~*~
+::Loofah::HTML5::SafeList::ACCEPTABLE_PROTOCOLS.add('apt')
```
dometto: We have a test that checks whether links can have the apt scheme. Also fine with removing that test.

lib/gollum-lib/sanitization.rb
```diff
@@ -1,378 +1,52 @@
...
-      if (add = env[:config][:add_attributes][node.name])
-        add.each do |key, value|
-          node[key] = value
-        end
-      end
+    def clean(data, historical = false)
+      doc = Loofah.fragment(data)
+      doc.scrub!(SCRUB_REMOVE)
+      doc.scrub!(:strip)
+      doc.scrub!(:nofollow) if historical
```
dometto: Loofah's built-in `:nofollow` scrubber allows a nice simplification of the 'history sanitization' logic.

lib/gollum-lib/sanitization.rb
```diff
@@ -1,378 +1,52 @@
...
-      'h4', 'h5', 'h6', 'hr', 'i', 'img', 'input', 'ins', 'kbd', 'label',
-      'legend', 'li', 'map', 'mark', 'math', 'menu', 'mfrac', 'mi', 'mn',
-      'mo', 'mrow', 'msqrt', 'msubsup', 'msup', 'mtext', 'ol', 'optgroup',
-      'option', 'p', 'pre', 'q', 's', 'samp', 'select', 'small', 'span',
-      'strike', 'strong', 'sub', 'sup', 'table', 'tbody', 'td', 'textarea',
-      'tfoot', 'th', 'thead', 'tr', 'tt', 'u', 'ul', 'var'
-    ] + MATHML_ELEMENTS).freeze
+    REMOVE_NODES = ['style', 'script']
 
+    SCRUB_REMOVE = Loofah::Scrubber.new do |node|
```
dometto: To explain: the `:strip` scrubber called in `#clean` removes unwhitelisted tags, but keeps their content. This is similar to Sanitize's behavior. However, we had some custom config in place that told Sanitize to complete remove some tags (i.e. also remove their content). This scrubber implements that behavior.

bartkamphorst: Nice!

lib/gollum-lib/sanitization.rb
bartkamphorst: 👍

dometto:
> PR is looking good. Should we wait out a little longer with merging though until we hear back about JRuby support?

👍

lib/gollum-lib/sanitization.rb
```diff
@@ -1,378 +1,52 @@
...
-      'mphantom', 'mroot', 'mrow', 'ms', 'mscarries', 'mscarry',
-      'msgroup', 'msline', 'mspace', 'msqrt', 'msrow', 'mstack',
-      'mstyle', 'msub', 'msup', 'msubsup', 'mtable', 'mtd',
-      'mtext', 'mtr', 'munder', 'munderover', 'semantics'
-    ].freeze
-
+    @@accepted_protocols = ::Loofah::HTML5::SafeList::ACCEPTABLE_PROTOCOLS.to_a.freeze
 
-    # Default whitelisted attributes required for MathML. See https://developer.mozilla.org/en-US/docs/Web/MathML/Attribute 
-    MATHML_ATTRS = {
```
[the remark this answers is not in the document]
dometto: A whole bunch of them! See [here](https://github.com/flavorjones/loofah/blob/master/lib/loofah/html5/safelist.rb#L145) I went and compared the two arrays and there are a couple of elements that don't overlap (some in ours that are not in loofah, some in loofah that are not in ours). I just assumed loofah's defaults are sane -- or maybe they have decided not to include some for safety concerns. Either way I'm happy to yield to their judgement.

Also notice that this will only affect mathml generated in the filter chain (e.g. by a rendering gem). Gollum's own support for mathml is implemented through JS: the mathml formulas are only converted to HTML after the page is loaded, and so loofah won't break any of that functionality.

lib/gollum-lib/filter/tags.rb
```diff
@@ -173,14 +173,8 @@ def parse_image_tag_options(options)
...
   def process_external_link_tag(url, pretty_name = nil)
-    accepted_protocols = @markup.wiki.sanitization.protocols['a']['href'].dup
-    if accepted_protocols.include?(:relative)
-      accepted_protocols.select!{|protocol| protocol != :relative}
-      regexp = %r{^((#{accepted_protocols.join("|")}):)?(//)}
-    else
-      regexp = %r{^((#{accepted_protocols.join("|")}):)}
-    end
-    if url =~ regexp
+    @accepted_protocols_regex ||= %r{^((#{::Gollum::Sanitization.accepted_protocols.join('|')}):)?(//)}
```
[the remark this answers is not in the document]
dometto: As far as I can see Sanitize had a separate flag for allowing `:relative`, loofah just accepts it.

What I don't really understand is the difference between the two regexes. Previously, if `:relative` was not allowed, the regex would match URI scheme's of the form `http:`, and not require `http://`. I don't know why the double slashes were not required in that case. Any thoughts?

lib/gollum-lib/page.rb
```diff
@@ -101,7 +101,7 @@ def name
     #
     # Returns the fully sanitized String title.
     def title
-      Sanitize.clean(name).strip
+      @wiki.sanitizer.clean(name).strip
```
dometto: Alternatively we could make a direct Loofah `:strip` or `:prune` call here (no need for the postprocessing in the `Sanitizer` class here).

dometto:
https://github.com/flavorjones/loofah/issues/88 is showing progress on getting loofah on JRuby to pass the test suite without any errors. I am inclined to merge this and switch to loofah. There are a couple of known errors, but that seems better to me than being tied to an old version of sanitize. What do you think, @bartkamphorst?

PR #325, opened 2019-03-22
[WIP] Goodbye Grit!

dometto:
This PR so far:
* Refactors the revert method (see [here](https://github.com/gollum/rugged_adapter/pull/33))
* Reactivates revert tests

Todo:
- [x] Wait until new rugged adapter release
- [x] Require the new version of rugged adapter / cleanup Gemfile
- [x] Reactive revert tests
- [x] Make sure the RJGit adapter also supports the new revert method
- [x] Remove the RJGit adapter reference in Gemfile, require the right version of it in `gemspec_java.rb`.
- [x] Set rugged as default adapter
- [x] Remove all references to grit

PR #352, opened 2019-11-14
Fix test for Page#versions

dometto:
Will be required after merging https://github.com/gollum/rugged_adapter/pull/39

Fortunately only one test failure to solve


dometto:
Think this PR ended up being unnecessary after taking a different approach in https://github.com/gollum/rugged_adapter/pull/39. Tests are passing in https://github.com/gollum/gollum-lib/pull/325/files anyway.

PR #356, opened 2020-03-09
Don't use kramdown when pandoc is available

dometto:
See https://github.com/gollum/gollum/issues/1470

### DIFF
--- a/lib/gollum-lib/markups.rb
+++ b/lib/gollum-lib/markups.rb
@@ -55,12 +55,16 @@ end
 
 module Gollum
   class Markup
-    GitHub::Markup::Markdown::MARKDOWN_GEMS['kramdown'] = proc { |content|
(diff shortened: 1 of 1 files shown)


dometto:
@bartkamphorst sanity check? This disables the use of kramdown when pandoc-ruby is present.

bartkamphorst:
Seems sane. 👍 Perhaps it would be even better if we not only check that pandoc-ruby is available, but actually test if pandoc is functioning. If not, fall back on kramdown.

dometto:
Yeah that would be nice, but not sure how to check for functioning. `gem_exists('pandoc-ruby')` is already doing this:
```ruby
      Gem::Specification.find {|spec| spec.name == name} != nil
```
I would say: if the pandoc-ruby gem is available, but not working, the fix is to uninstall it. github-markup also doesn't check whether a rendering gem is actually functional; if you have, say, commonmark installed but it's somehow broken, it won't fall back to the next gem (I think). Is it acceptable to merge this as is?

bartkamphorst:
I was thinking that the gem `pandoc-ruby` could be installed fine, but that pandoc itself could be dysfunctional. You are right though that there are other gems that run the same risk. This is fine to merge as is!

PR #358, opened 2020-03-16
Restore Java gemfile

dometto:
I removed the separate Java gemspec because the rjgit_adapter is already set as default in the 'normal' `gollum-lib.gemspec`. However, I just realized the reason for having the second gemspec is to facilitate building the java version of the gem -- even though in normal cases, the dependency chain is resolved through `gollum-lib.gemspec`. So I've re-added the Java gemspec.

What annoys me, however, is that we have to keep track of the rjgit_adapter dependency in two places: in `gollum-lib.gemspec`, and `gollum-lib_java.gemspec`. The only way I can see around that is have a *third* file, `adapter_dependencies.rb` or some such, in which we define the requirements, and load that file from both gemspec files. Does that seem better or worse?

### DIFF
--- a/adapter_dependencies.rb
+++ b/adapter_dependencies.rb
@@ -1,0 +1,5 @@
+if RUBY_PLATFORM == 'java' then
+  DEFAULT_ADAPTER_REQ = ['gollum-rjgit_adapter', '>= 0.5.1', '~> 0.5.1']
+else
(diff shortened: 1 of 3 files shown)


bartkamphorst:
> The only way I can see around that is have a third file, adapter_dependencies.rb or some such, in which we define the requirements, and load that file from both gemspec files. Does that seem better or worse?

I don't really mind either way tbh.

dometto:
Went for a separate file so we don't run the risk of updating the dependency in only one place.

bartkamphorst:
Looks solid.

PR #359, opened 2020-03-16
Fix titles

dometto:
It turns out that when rendering markdown with Pandoc, the result was a fully formed HTML document, rather than just the body's inner html, which is what we require. This was causing some of the Nokogiri operations we do elsewhere in the filter chain and in gollum to fail, resulting (perhaps amongst other bugs) in the bug that h1 title's weren't being found when using Pandoc (see https://github.com/gollum/gollum/issues/1484).

This PR remedies this by extracting the body from the document that pandoc renders.


dometto:
Thanks, works like a charm!

PR #360, opened 2020-03-16
Look for linked pages in same directory first

dometto:


dometto:
If you link `[[Bar]]` in `greek/Foo`, it will first check for `greek/Bar`, and if it doesn't exist, try `./Bar`.

dometto:
@bartkamphorst if tests pass and this doesn't require modifications, feel free to merge -- I have to run now :)

test/test_markup.rb
```diff
@@ -741,6 +741,17 @@
     output = Gollum::Markup.new(page).render
     assert_html_equal %{<p>a <a href="/alpha.jpg">Alpha</a> b</p>}, output
   end
+  
+  test "page link with relative path" do
```
bartkamphorst: Shouldn't we also test the other case, namely where `LinkedRelative.md` lives in the root?

lib/gollum-lib/filter/tags.rb
```diff
@@ -236,7 +236,11 @@ def process_page_link_tag(link_part, pretty_name = nil)
   #
   # Returns a Gollum::Page instance if a page is found, or nil otherwise
   def find_page_from_path(path)
-    @markup.wiki.page(path)
+    if !Pathname.new(path).absolute? && result = @markup.wiki.page(::File.join(@markup.dir, path))
```
bartkamphorst: I prefer `if Pathname.new(path).relative? && [...]`.

bartkamphorst:
> If you link [[Bar]] in greek/Foo, it will first check for greek/Bar, and if it doesn't exist, try ./Bar.

To clarify: do you mean:
If you link [[Bar]] in **./greek/Foo** , it will first check for **./greek/Bar**, and if it doesn't exist, try **/Bar**?

dometto:
>  If you link [[Bar]] in ./greek/Foo , it will first check for ./greek/Bar, and if it doesn't exist, try /Bar?

Yes!

dometto:
Possible behavior would also be to not fall back to checking **/Bar**, so that absolute paths have to be specified explicitly.

bartkamphorst:
Right. Okay, so I'm now thinking the second step shouldn't happen, at least not by default.

dometto:
> Right. Okay, so I'm now thinking the second step shouldn't happen, at least not by default.

Implemented this behavior. Also see the new text for how link text should behave.

Note the following: the link text can either be a file path, in which we case we want to call `File.basename` on it, or a user-defined link text (when you do e.g. `[[Foo|My link text]]``, in which case we don't want to call File.basename on it. I'm using `str.include?('/')` as a heuristic for determining whether `str` is a filepath. This means people can't define link texts that include a slash, though . :-(

dometto:
* Link texts for existing pages will be of the following form `[[Subdir/Foo.md]]` -> `Foo`
* Link texts for non-existing pages will default to the file path: `[[Subdir/Doesntexist]]` -> `Subdir/Doesntexist`

lib/gollum-lib/filter/tags.rb
```diff
@@ -189,6 +188,7 @@ def process_external_link_tag(url, pretty_name = nil)
   # Returns the String HTML if the tag is a valid file link tag or nil
   #   if it is not.
   def process_file_link_tag(link_part, pretty_name)
+    return nil if ::Gollum::Page.valid_extension?(link_part)
```
dometto: Managed to catch a nasty edge case here when I added one of the tests for link names. If you'd specify a link to a page including its file extension, the result would be a **file** link, not an **absent page** link, because `process_file_link_tag` would succeed (since all Pages are also Files).

bartkamphorst: This is when the page in question does not exist, correct?

test/test_markup.rb
```diff
@@ -125,14 +125,22 @@
     assert_equal 1, anchors.size
     assert_equal 'internal absent', anchors[0]['class']
     assert_equal '/Page', anchors[0]['href']
-    assert_equal 'Page', anchors[0].text
+    assert_equal '/Page', anchors[0].text
   end
-  
+    
   test "absolute link link text" do
```
bartkamphorst: test "link text of absolute link" reads a little better.

test/test_markup.rb
```diff
@@ -125,14 +125,22 @@
...
+    @wiki.write_page("Docs/Integration/How the future will look", :markdown, "Bright", commit_details)
+    @wiki.write_page("linktexttest", :markdown, "[[Docs/Integration/How the future will look.md]]", commit_details)
+    page   = @wiki.page("linktexttest")
+    output = Gollum::Markup.new(page).render
+    assert_html_equal %{<p><a class="internal present" href="/Docs/Integration/How%20the%20future%20will%20look.md">How the future will look</a></p>}, output
+  end
+  
+  test "broken absolute link link text" do
```
bartkamphorst: test "link test of broken absolute link".

lib/gollum-lib/filter/tags.rb
dometto: It's a little more complicated unfortunately. At this point in the chain, we have not yet tested whether the page exists. We need to test for a matching **File** before we test for a matching **Page**, because otherwise we could never get the result that the link is an "internal absent" link. If there's not a Page, there could yet be a file, so we can't mark it as absent on that. But File links aren't marked as absent or present at all.

So what this does is: it stops the tag filter from treating something as a link to a File if it has a valid Page extension. Then we go on to evaluate the link as a possible Page extension, and we check for Page existence.

An alternative to this could be to completely eradicate the difference between Page and File links: just look for a File in all cases, and mark the link as "internal present" or "internal absent".

dometto: > An alternative to this could be to completely eradicate the difference between Page and File links: just look for a File in all cases, and mark the link as "internal present" or "internal absent".

Just realized this is not possible, as the file-finding algorithm doesn't have the required "slack" for matching page extensions to actually find any pages. :(

Hope this fix seems ok!

dometto:
Not sure you got to see this message, as I think I accidentally posted in a Review I didn't submit:

An alternative to [this](https://github.com/gollum/gollum-lib/pull/360/files/b8d0f3dc7d322bb47065c1ea10e129b263fe1d16#diff-9c66db1000504e8fd86544f23961803eR191) could be to completely eradicate the difference between Page and File links: just look for a File in all cases, and mark the link as "internal present" or "internal absent".

Just realized this is not possible, as the file-finding algorithm doesn't have the required "slack" for matching page extensions to actually find any pages. :(

Hope this fix seems ok!

lib/gollum-lib/helpers.rb
```diff
@@ -1,13 +1,23 @@
...
+      return nil if url.nil?
       url.gsub!('%2F', '/')
       return '/' + url.gsub(/^\/+/, '') if url[0, 1] == '/'
       url
     end
+    
+    # Take a path and turn it into a string for display as link text.
+    # For example:
+    # '/opt/local/bin/ruby.ext' -> 'ruby'
+    def path_to_link_text(str, is_path = true)
```
bartkamphorst: What's a little confusing is that the comment states that this methods takes a path, but the method signature says it takes a String `str` that may or may not be a path.

dometto: Good point! I can just change the comment, unless you see a better way of implementing the logic. See my comment in `tags.rb`

lib/gollum-lib/filter/tags.rb
```diff
@@ -225,7 +225,7 @@ def process_page_link_tag(link_part, pretty_name = nil)
     end
     presence = :page_present if page
 
-    name = pretty_name ? pretty_name : link
+    name = pretty_name ? pretty_name : path_to_link_text(link, !!page)
```
dometto: I added the second argument to `path_to_link_text` because I couldn't see a good way of writing it here. The logic needs to be this:

```ruby
if page
  name =  name = pretty_name ? pretty_name : path_to_link_text(link)
  link = page.escaped_url_path
else
  name =  name = pretty_name ? pretty_name : link
  link = ERB::Util.url_encode(link).force_encoding('utf-8')
end
```

But as you can see that involves some annoying duplication

lib/gollum-lib/helpers.rb
bartkamphorst: And what about changing `is_path` to `for_page`? You are passing in `!!page` after all.

dometto: 👍

PR #363, opened 2020-03-21
Add compatibility option for 4.x style links

dometto:


lib/gollum-lib/filter/tags.rb
```diff
@@ -241,7 +241,11 @@ def process_page_link_tag(link_part, pretty_name = nil)
   # Returns a Gollum::Page instance if a page is found, or nil otherwise
   def find_page_from_path(path)
     if Pathname.new(path).relative?
-      @markup.wiki.page(::File.join(@markup.dir, path))
+      page = @markup.wiki.page(::File.join(@markup.dir, path))
+      if page.nil? && @markup.wiki.link_compatibility # 4.x link compatibility option. Slow!
+        page = @markup.wiki.pages.detect {|page| page.path =~ /^(.*\/)?#{path}\..+/i}
```
dometto: This is slow because it looks up all the wiki's pages first. A faster way of implementing this would have to make changes to `Page#path_match`.

dometto:
More speedy implementation now.

lib/gollum-lib/filter/tags.rb
```diff
@@ -241,7 +241,11 @@ def process_page_link_tag(link_part, pretty_name = nil)
   # Returns a Gollum::Page instance if a page is found, or nil otherwise
   def find_page_from_path(path)
     if Pathname.new(path).relative?
-      @markup.wiki.page(::File.join(@markup.dir, path))
+      page = @markup.wiki.page(::File.join(@markup.dir, path))
+      if page.nil? && @markup.wiki.link_compatibility # 4.x link compatibility option. Slow!
+        page = @markup.wiki.page(path, nil, true)
```
dometto: If we haven't found the page in the subdir, do `wiki.page` with `global_match` set to true. @bartkamphorst do you think this behavior should be default, or that we should have the `--link-compatibility` toggle?

bartkamphorst: Let's make this the default behavior, at least for now.

PR #365, opened 2020-03-24
Github compatibility option

dometto:
This feature enables a 'fallback' option: when linking `[[Bilbo Baggins]]` and it is not found, look for `[[Bilbo-Baggins]]`. Provides backwards compatibility with 4.x and github. Questions:

* do we want this?
* if so, do we want this and `global_tag_lookup` to be different options, or the same?
   * could also make them separate options in gollum-lib, but have one `--tag-compatibility' (? name) switch in gollum which enables them both.

### DIFF
--- a/lib/gollum-lib/file.rb
+++ b/lib/gollum-lib/file.rb
@@ -12,7 +12,7 @@ module Gollum
       #
       # query     - The String path to match.
       # entry     - The BlobEntry to check against.
-      # global_match - If true, find a File matching path's filename, but not it's directory (so anywhere in the repo)
+      # global_match - If true, find a File matching path's filename, but not its directory (so anywhere in the repo)
       def path_match(query, entry, global_match = false)
         query == ::File.join('/', entry.path)
       end
--- a/lib/gollum-lib/file.rb
+++ b/lib/gollum-lib/file.rb
@@ -12,8 +12,8 @@ module Gollum
       #
       # query     - The String path to match.
       # entry     - The BlobEntry to check against.
-      # global_match - If true, find a File matching path's filename, but not it's directory (so anywhere in the repo)
-      # sub_spaces   - GitHub compatibility: substitutes - for spaces when comparing filenames.
+      # global_match - (Not implemented for File) If true, find a File matching path's filename, but not it's directory (so anywhere in the repo)
+      # sub_spaces   - (Not implemented for File) Substitutes spaces for - when comparing filenames.
       def path_match(query, entry, global_match = false, sub_spaces = false)
         query == ::File.join('/', entry.path)
       end
--- a/lib/gollum-lib/file.rb
+++ b/lib/gollum-lib/file.rb
@@ -12,7 +12,9 @@ module Gollum
       #
       # query     - The String path to match.
       # entry     - The BlobEntry to check against.
-      def path_match(query, entry, global_match = false)
+      # global_match - If true, find a File matching path's filename, but not it's directory (so anywhere in the repo)
+      # sub_spaces   - GitHub compatibility: substitutes - for spaces when comparing filenames.
+      def path_match(query, entry, global_match = false, sub_spaces = false)
         query == ::File.join('/', entry.path)
       end
     end
@@ -24,11 +26,13 @@ module Gollum
     # version - The String version ID to find.
     # try_on_disk - If true, try to return just a reference to a file
     #               that exists on the disk.
+    # global_match - If true, find a File matching path's filename, but not it's directory (so anywhere in the repo)
+    # sub_spaces   - GitHub compatibility: substitutes - for spaces when comparing filenames.
     #
     # Returns a Gollum::File or nil if the file could not be found. Note
     # that if you specify try_on_disk=true, you may or may not get a file
     # for which on_disk? is actually true.
-    def self.find(wiki, path, version, try_on_disk = false, global_match = false)
+    def self.find(wiki, path, version, try_on_disk = false, global_match = false, sub_spaces = false)
       map = wiki.tree_map_for(version.to_s)
 
       query_path = Pathname.new(::File.join(['/', wiki.page_file_dir, path].compact)).cleanpath.to_s
@@ -36,7 +40,7 @@ module Gollum
 
       begin
         entry = map.detect do |entry|
-          path_match(query_path, entry, global_match)
+          path_match(query_path, entry, global_match, sub_spaces)
         end
         entry ? self.new(wiki, entry.blob(wiki.repo), entry.dir, version, try_on_disk) : nil
       rescue Gollum::Git::NoSuchShaFound
--- a/lib/gollum-lib/file.rb
+++ b/lib/gollum-lib/file.rb
@@ -12,7 +12,7 @@ module Gollum
       #
       # query     - The String path to match.
       # entry     - The BlobEntry to check against.
-      # global_match - (Not implemented for File) If true, find a File matching path's filename, but not it's directory (so anywhere in the repo)
+      # global_match - If true, find a File matching path's filename, but not it's directory (so anywhere in the repo)
       def path_match(query, entry, global_match = false)
         query == ::File.join('/', entry.path)
       end
--- a/lib/gollum-lib/file.rb
+++ b/lib/gollum-lib/file.rb
@@ -13,8 +13,7 @@ module Gollum
       # query     - The String path to match.
       # entry     - The BlobEntry to check against.
       # global_match - (Not implemented for File) If true, find a File matching path's filename, but not it's directory (so anywhere in the repo)
-      # sub_spaces   - (Not implemented for File) Substitutes spaces for - when comparing filenames.
-      def path_match(query, entry, global_match = false, sub_spaces = false)
+      def path_match(query, entry, global_match = false)
         query == ::File.join('/', entry.path)
       end
     end
@@ -27,12 +26,11 @@ module Gollum
     # try_on_disk - If true, try to return just a reference to a file
     #               that exists on the disk.
     # global_match - If true, find a File matching path's filename, but not it's directory (so anywhere in the repo)
-    # sub_spaces   - GitHub compatibility: substitutes - for spaces when comparing filenames.
     #
     # Returns a Gollum::File or nil if the file could not be found. Note
     # that if you specify try_on_disk=true, you may or may not get a file
     # for which on_disk? is actually true.
-    def self.find(wiki, path, version, try_on_disk = false, global_match = false, sub_spaces = false)
+    def self.find(wiki, path, version, try_on_disk = false, global_match = false)
       map = wiki.tree_map_for(version.to_s)
 
       query_path = Pathname.new(::File.join(['/', wiki.page_file_dir, path].compact)).cleanpath.to_s
@@ -40,7 +38,7 @@ module Gollum
 
       begin
         entry = map.detect do |entry|
-          path_match(query_path, entry, global_match, sub_spaces)
+          path_match(query_path, entry, global_match)
         end
         entry ? self.new(wiki, entry.blob(wiki.repo), entry.dir, version, try_on_disk) : nil
       rescue Gollum::Git::NoSuchShaFound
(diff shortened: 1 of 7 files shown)


lib/gollum-lib/file.rb
```diff
@@ -12,7 +12,9 @@ class << self
       #
       # query     - The String path to match.
       # entry     - The BlobEntry to check against.
-      def path_match(query, entry, global_match = false)
+      # global_match - If true, find a File matching path's filename, but not it's directory (so anywhere in the repo)
+      # sub_spaces   - GitHub compatibility: substitutes - for spaces when comparing filenames.
+      def path_match(query, entry, global_match = false, sub_spaces = false)
         query == ::File.join('/', entry.path)
```
dometto: global_match and sub_spaces not implemented for Files, which I think is ok?

lib/gollum-lib/page.rb
```diff
@@ -10,11 +10,18 @@ class << self
...
-      def path_match(query, entry, global_match = false)
+      # global_match - If true, find a File matching path's filename, but not it's directory (so anywhere in the repo)
+      # sub_spaces   - GitHub compatibility: substitutes - for spaces when comparing filenames.
+      def path_match(query, entry, global_match = false, sub_spaces = false)
         return false if "#{entry.name}".empty?
         return false unless valid_extension?(entry.name)
-        entry_name = valid_extension?(query) ? entry.name : strip_filename(entry.name)        
-        query == (global_match ? ::File.join('/', entry_name) : ::File.join('/', entry.dir, entry_name))
+        entry_name = valid_extension?(query) ? entry.name : strip_filename(entry.name)     
+        match_path = ::File.join([
```
dometto: This refactor is better than line 17 in the original anyway, even if we don't want the `sub_spaces` options.

lib/gollum-lib/page.rb
```diff
@@ -10,11 +10,18 @@ class << self
...
+          '/',
+          global_match ? nil : entry.dir,
+          sub_spaces ? entry_name.gsub(' ', '-') : entry_name
```
dometto: `gsub` spaces, or `\s`?

lib/gollum-lib/file.rb
bartkamphorst: Depends on github's behavior, and 4.x behavior I suppose?

lib/gollum-lib/page.rb
bartkamphorst: 👍

lib/gollum-lib/wiki.rb
```diff
@@ -93,8 +93,8 @@ def default_options
     #           :mathjax       - Set to false to disable mathjax.
     #           :user_icons    - Enable user icons on the history page. [gravatar, identicon, none].
     #                            Default: none
-    #           :show_all      - Show all files in file view, not just valid pages.
-    #                            Default: false
+    #           :global_tag_lookup        - Enable 4.x compatibility behavior for links
+    #           :github_tag_compatibility - Enable github compatibility behavior for links (substitute - for spaces in page names)
```
bartkamphorst: `substitute spaces by hyphens (-) in page names`

dometto:
@petrkalina if you can manage running gollum using this gollum-lib branch from source, you could see if it helps with your use case. Steps:

1. Add `gem 'gollum-lib', :git => 'https://github.com/gollum/gollum-lib/pull/365', :branch => 'github_tags'` to your `Gemfile`
2. run `bundle install` (or `bundle update gollum-lib`, I think)
3. Run gollum with a `config.rb`:

```ruby
Precious::App.settings.wiki_options[github_tag_compatibility] = true
```

lib/gollum-lib/wiki.rb
bartkamphorst: I would prefer having a name that does not include github, as gollum is also used by other organizations (notably gitlab). Something like `hyphenate-tags` perhaps or `hyphenated-tag-lookup`?

test/test_markup.rb
```diff
@@ -769,6 +769,19 @@
     assert_html_equal %{<p>a <a class="internal present" href="/LinkedRelative.md">LinkedRelative</a> b</p>}, output 
   end
   
+  test "page link with github compatibility" do
```
bartkamphorst: Same note as before about explicitly naming github.

lib/gollum-lib/wiki.rb
dometto: Ah yes, much better! Or `hyphened-tag-lookup` as its slightly simpler?

petrkalina:
> @petrkalina if you can manage running gollum using this gollum-lib branch from source, you could see if it helps with your use case. Steps:
> 
> 1. Add `gem 'gollum-lib', :git => 'https://github.com/gollum/gollum-lib/pull/365', :branch => 'github_tags'` to your `Gemfile`
> 2. run `bundle install` (or `bundle update gollum-lib`, I think)
> 3. Run gollum with a `config.rb`:
> 
> ```ruby
> Precious::App.settings.wiki_options[:github_tag_compatibility] = true
> ```

I'm running from `5.x` branch of gollum. I'm note yet sure how exactly does the source of gollum-lib refer to gollum. What I do is checkout `5.x` branch of gollum and then run the following Dockerfile

```
FROM ruby

RUN apt-get -y update \
    && apt-get -y install cmake \
    && apt-get -y install nodejs \
    && rm -rf /var/lib/apt/lists/*

RUN gem install github-markdown

COPY gollum /gollum
RUN bundle install --gemfile gollum/Gemfile

WORKDIR /gollum

ENTRYPOINT ["bundle", "exec", "bin/gollum", "--port", "80", "--allow-uploads", "page", "/wiki"]
EXPOSE 80
```

i see:

```
Step 5/8 : RUN bundle install --gemfile gollum/Gemfile
 ---> Running in ba683d921767
Fetching https://github.com/gollum/gollum-lib.git
...
Using gollum-lib 5.0.a.4 from https://github.com/gollum/gollum-lib.git (at gollum-lib-5.x@d9ffb7f)
F
```

which suggests that I'm using the correct branch. I can run the migrate script by i.e.:

```
root@gollum-v5-1:/gollum# bundle exec bin/gollum-migrate-tags --help
Use this tool to migrate a wiki repository created under Gollum versions 4.x or earlier to a 5.x compatibile repo.
```

is this all correct?

lib/gollum-lib/file.rb
dometto: On 4.x both global_match and sub_spaces work for Files, too. But Github wikis at present doesn't allow uploading Files anyway -- you can only add pages. Would it be an acceptable divergence from 4.x not to implement compatibility for this?

dometto:
@petrkalina yes that's correct, but you need to modify your `gollum/Gemfile` in accordance with my 1. above (add this branch of gollum-lib to it) to use the compatibility option.

lib/gollum-lib/wiki.rb
bartkamphorst: Yep, fine.

lib/gollum-lib/file.rb
bartkamphorst: I'm now thinking we should also add compatibility for Files, but only if it's doable.

dometto: Done in the refactor I've just pushed

petrkalina:
> @petrkalina yes that's correct, but you need to modify your `gollum/Gemfile` in accordance with my 1. above (add this branch of gollum-lib to it) to use the compatibility option.

I modified the Gemfile to

```
source 'https://rubygems.org'

if RUBY_PLATFORM == 'java'
  gem 'warbler'
end

gem 'gollum-lib', :git => 'https://github.com/gollum/gollum-lib/pr/365', :branch => 'github_tags'
gem 'therubyrhino', :platforms => :jruby
gemspec

gem "rake", '~> 12.3', '>= 12.3.3'
```

but I'm getting

```
root@008da3d66f52:/gollum# bundle install                          
Fetching https://github.com/gollum/gollum-lib/pr/365
remote: Not Found
fatal: repository 'https://github.com/gollum/gollum-lib/pr/365/' not found
```

I never checked out github pull requests before and I never used bundler before so sorry if I'm missing anything obvious..

dometto:
Sorry, my bad! It should be:

`gem 'gollum-lib', :git => 'https://github.com/gollum/gollum-lib', :branch => 'github_tags'`

Also, I've just renamed the config option. It should now be:

```ruby
Precious::App.settings.wiki_options[:hyphened_tag_lookup] = true
```

dometto:
@bartkamphorst took a simpler approach now which also applies to Files :)

petrkalina:
> Sorry, my bad! It should be:
> 
> `gem 'gollum-lib', :git => 'https://github.com/gollum/gollum-lib', :branch => 'github_tags'`
> 

now I get:

```
root@008da3d66f52:/gollum# bundle install
Fetching https://github.com/gollum/gollum-lib
fatal: Needed a single revision
Revision github_tags does not exist in the repository https://github.com/gollum/gollum-lib. Maybe you misspelled it?
```

dometto:
Argh. Sorry again, doing too much at the same time. Should be:

`gem 'gollum-lib', :git => 'https://github.com/repotag/gollum-lib', :branch => 'github_tags'`

petrkalina:
> Argh. Sorry again, doing too much at the same time. Should be:
> 
> `gem 'gollum-lib', :git => 'https://github.com/repotag/gollum-lib', :branch => 'github_tags'`

I just figured that out and it was not easy :-)! Thanks ..

dometto:
@petrkalina have you managed to test this, and if so, does it resolve your issue?

lib/gollum-lib/file.rb
```diff
@@ -24,6 +25,7 @@ def path_match(query, entry, global_match = false)
     # version - The String version ID to find.
     # try_on_disk - If true, try to return just a reference to a file
     #               that exists on the disk.
+    # global_match - If true, find a File matching path's filename, but not it's directory (so anywhere in the repo)
```
bartkamphorst: `its directory`

lib/gollum-lib/file.rb
```diff
@@ -12,6 +12,7 @@ class << self
       #
       # query     - The String path to match.
       # entry     - The BlobEntry to check against.
+      # global_match - If true, find a File matching path's filename, but not it's directory (so anywhere in the repo)
```
bartkamphorst: `its directory`

lib/gollum-lib/filter/tags.rb
```diff
@@ -239,15 +239,16 @@ def process_page_link_tag(link_part, pretty_name = nil)
...
   # Returns a Gollum::Page instance if a page is found, or nil otherwise
-  def find_page_from_path(path)
-    if Pathname.new(path).relative?
-      page = @markup.wiki.page(::File.join(@markup.dir, path))
-      if page.nil? && @markup.wiki.global_tag_lookup # 4.x link compatibility option. Slow!
-        page = @markup.wiki.page(path, nil, true)
+  def find_page_or_file_from_path(path, kind = :page)
+    query = query_for_path(path)
+    if Pathname.new(query).relative?
+      result = @markup.wiki.send(kind, ::File.join(@markup.dir, query))
```
bartkamphorst: Nice. 👍

lib/gollum-lib/filter/tags.rb
```diff
@@ -357,4 +358,8 @@ def generate_image_attributes(options)
 
     return classes, attrs, containered
   end
+  
+  def query_for_path(path)
+    @markup.wiki.hyphened_tag_lookup ? path.gsub(' ', '-') : path
```
bartkamphorst: Just to be clear: we don't want a fallback construction here, correct? That if no page/file is found with spaces that we look for a hyphenated version. Seems like that would be a lot of overhead, but just checking.

dometto: Yes, exactly. And it would raise the question of how to combine that fallback with the `global_tag_lookup`: one fallback search that enables both `global_tag_lookup` and `hyphened_tag_lookup`? Two fallback searches? Etc. So think it's best to just enable this from the very first search on. Means that with `--hyphened-tag-lookup` people cannot use spaces in filenames at all -- but that seems to be the ask.

dometto:
Thanks @bartkamphorst!

lib/gollum-lib/filter/tags.rb
bartkamphorst: OK!

dometto:
@petrkalina different approach: I'll just merge this and https://github.com/gollum/gollum/pull/1497  so you can test the new compatibility behavior with a fresh install of `5.x` and the `--hyphened-tag-lookup` command line option.

PR #366, opened 2020-03-26
Case insentive lookup option, and fix hyphened lookup option

dometto:


lib/gollum-lib/file.rb
```diff
@@ -12,9 +12,18 @@ class << self
       #
       # query     - The String path to match.
       # entry     - The BlobEntry to check against.
-      # global_match - If true, find a File matching path's filename, but not its directory (so anywhere in the repo)
-      def path_match(query, entry, global_match = false)
-        query == ::File.join('/', entry.path)
+      # global_match - (Not implemented for File) If true, find a File matching path's filename, but not its directory (so anywhere in the repo)
```
bartkamphorst: This comment is confusing: `(Not implemented for File) If true, find a File`

lib/gollum-lib/file.rb
```diff
@@ -12,9 +12,18 @@ class << self
...
+      # hyphened_tags  - If true, replace spaces in match_path with hyphens.
+      # case_insensitive - If true, compare query and match_path case-insensitively
+      def path_match(query, entry, global_match = false, hyphened_tags = false, case_insensitive = false)
+        path_compare(query, ::File.join('/', entry.path), hyphened_tags, case_insensitive)
+      end
+      
+      # For use with self.path_match: returns true if 'query' and 'match_path' match, strictly or taking account of the following parameters:
+      # hyphened_tags  - If true, replace spaces in match_path with hyphens.
+      # case_insensitive - If true, compare query and match_path case-insensitively
+      def path_compare(query, match_path, hyphened_tags, case_insensitive = false)
```
bartkamphorst: Why is `hyphened_tags` mandatory but `case_insensitive` optional here in `path_compare`?

lib/gollum-lib/wiki.rb
```diff
@@ -95,6 +95,7 @@ def default_options
     #                            Default: none
     #           :global_tag_lookup        - Enable 4.x compatibility behavior for links
     #           :hyphened_tag_lookup - Spaces in tag paths are treated as dashes (-)
+    #           :case_insensitive_tag_lookup - Paths in tags are compared case-case_insensitively
```
bartkamphorst: `case-case_insensitively` typo?

lib/gollum-lib/file.rb
dometto: That's a mistake, thanks -- they should both be mandatory here.

lib/gollum-lib/wiki.rb
dometto: Indeed!

PR #367, opened 2020-03-27
Better GitHub tags

dometto:


dometto:
Accidentally merged #367 before pushing my final local commits which addressed @bartkamphorst's questions.

PR #370, opened 2020-03-31
Fix redirects when there is no @wiki.repo.head

dometto:
See https://github.com/gollum/gollum/issues/1530

I'll add a test to gollum checking for an OK response when using en empty repository.

### DIFF
--- a/lib/gollum-lib/redirects.rb
+++ b/lib/gollum-lib/redirects.rb
@@ -7,12 +7,12 @@ module Gollum
   module Redirects
         
     def stale?
-      @current_head != @wiki.repo.head.commit.sha
+      @current_head != get_head_sha
     end
 
     def init(wiki)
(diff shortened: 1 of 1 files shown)


lib/gollum-lib/redirects.rb
```diff
@@ -33,6 +33,10 @@ def dump
       @wiki.overwrite_file(REDIRECTS_FILE, self.to_yaml, {})
     end
     
+    def get_head_sha
+      @wiki.repo.head.nil? ? nil : @wiki.repo.head.commit.sha
```
bartkamphorst: Slightly more efficient by not checking explicitly for nil:
`@wiki.repo.head ? @wiki.repo.head.commit.sha : nil`

dometto: Thanks!

PR #371, opened 2020-04-01
Add base_path to /gollum/emoji

heavywatal:
Briefly confirmed that this PR does not break the behavior when `base_path` is `Nil`.

### DIFF
--- a/lib/gollum-lib/filter/emoji.rb
+++ b/lib/gollum-lib/filter/emoji.rb
@@ -26,7 +26,8 @@ class Gollum::Filter::Emoji < Gollum::Filter
   end
 
   def process(data)
-    data.gsub! process_pattern, %q(<img src="/gollum/emoji/\k<name>" alt="\k<name>" class="emoji">)
+    src = ::File.join(@markup.wiki.base_path, '/gollum/emoji/\\k<name>')
+    data.gsub! process_pattern, %Q(<img src="#{src}" alt="\\k<name>" class="emoji">)
     data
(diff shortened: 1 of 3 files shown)


dometto:
Great find, many thanks! Unfortunately, this is now causing a test error. You could fix this by changing the `MockWiki` class here:

https://github.com/gollum/gollum-lib/blob/master/test/helper.rb#L71

To set the `@base_path = '/'`, and add an `attr_reader :base_path` [here](https://github.com/gollum/gollum-lib/blob/master/test/helper.rb#L86). 

Moreover, it would be great to have a test guarding against this in https://github.com/gollum/gollum-lib/blob/master/test/filter/test_emoji.rb! Would you be willing to add that?

> Briefly confirmed that this PR does not break the behavior when base_path is Nil.

This is because `base_path` is set to `'/`` [here](https://github.com/gollum/gollum-lib/blob/master/lib/gollum-lib/wiki.rb#L128). Good enough, I think!

heavywatal:
Thank you for the instruction. A quick fix submitted. Next, I will try adding test.

dometto:
looks good to me! For some reason Travis is not running the tests for the latest commit, so I'll close and reopen this PR.

dometto:
Thanks @heavywatal! I'll wait a day or two before doing a bugfix release, to see if anything else comes up, unless it is critical for you that this is released ASAP.

heavywatal:
No problem!

PR #378, opened 2020-05-14
Make default extension configurable

dometto:


dometto:
See https://github.com/gollum/gollum/issues/1561

PR #380, opened 2020-06-06
Use the currently active page name in Series macro instead of subpage name

### LINKED ISSUE #379: Series macro can't be used in subpages
Using the [Series](https://github.com/gollum/gollum/wiki/Standard-Macros#series) macro in subpages looks like it tries to match the name of the subpage to the prefix instead of the current page, then fails giving an error: `Macro Error for Series: This page's name does not match the prefix '<prefix>'`.

I've had a look and will open a PR but I've never used Ruby before so might need a bit of nudging in the direction of best practice.

tpoliaw:
Fixes #379. I've not used Ruby before so happy to rewrite if there's a better way of doing it or a better place for the change.

### DIFF
--- a/lib/gollum-lib/macro.rb
+++ b/lib/gollum-lib/macro.rb
@@ -25,6 +25,10 @@ module Gollum
       "<p class=\"gollum-error\">#{s}</p>"
     end
 
+    def active_page
+      return @page.parent_page || @page
+    end
+
     # The special class we reserve for only the finest of screwups.  The
     # underscore is to make sure nobody can define a real, callable macro
     # with the same name, because that would be... exciting.
--- a/lib/gollum-lib/macro/series.rb
+++ b/lib/gollum-lib/macro/series.rb
@@ -3,7 +3,7 @@ module Gollum
 
     class Series < Gollum::Macro
       def render(series_prefix = "")
-      	raise "This page's name does not match the prefix '#{series_prefix}'" unless @page.name =~ /^#{series_prefix}/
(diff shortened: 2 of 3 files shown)


dometto:
@tpoliaw many thanks for spotting the bug and for this PR! The direction of this solution is fine, but we have a minor comment. We think the concept of an "active page" is not a very clear when implemented under the `Page` class: Pages are active in the frontend, but in the backend it can be confusing. So we'd like to propose that instead, you define an `active_page` method in the `Series` class:

```ruby
def active_page
  @page.parent_page&.name || @page.name
end
```

Does that seem reasonable?

tpoliaw:
Sure, that makes sense. Is it something that would be useful for other macros as well or just series?

One comment on your suggested function, would it be better to have either `active_page` that returns the page or `active_name` that returns the name instead of a mix?

dometto:
`active_page` (so returning page, not name) could potentially be useful for other Macros, so you could add that to the Macro class I guess, and then use `active_page.name` in the `Series` Macro! Thanks for your work!

lib/gollum-lib/macro.rb
```diff
@@ -25,6 +25,10 @@ def html_error(s)
       "<p class=\"gollum-error\">#{s}</p>"
     end
 
+    def active_page()
```
dometto, line 28: just for stylistic unity, would you mind removing the parantheses here? Ruby doesn't need them :) After that I'll merge this PR!

dometto:
Thanks @tpoliaw, and sorry for the long wait before merging!

PR #381, opened 2020-06-11
Only remove valid extensions when generating link display part

dometto:
See https://github.com/gollum/gollum/issues/1570#issuecomment-642471871

Previously, the display part of a link `[[v1.2]]` would be `v1`, because `.2` was being treated as a file extension, and thus removed. This PR only chops off the extension if it's a valid page extension.


bartkamphorst:
Seems sensible, @dometto . 👍

tevino:
@dometto Will this be released soon?

dometto:
Sorry for the long wait @tevino. Just released!

tevino:
Thank you @dometto

PR #382, opened 2020-06-25
add tree_toc macro

acavalin:
The TreeTOC macro generates an indented file&folders list like a filemanager.

It can be styled to obtain a nice TOC:
sample styled tree


dometto:
Closing because this lags behind development by now. Apologies for the slow response and thanks for chipping in. Sometimes we're not sure how to best integrate a PR given the direction the project is moving in, and given the limited time at our disposal to work on gollum, that sometimes mean we can't respond to everything. Thanks for chipping in though, and please don't hesitate to contribute in the future!

PR #383, opened 2020-07-08
Make pandoc compatible with MathJax

ashwinreddy:
This PR came out of the discussion on https://github.com/gollum/gollum/issues/1589. This change should allow people to install pandoc ruby without seeing any changes to their LaTeX code rendered by MathJax. To do this, we have to tell pandoc-ruby not to treat TeX in any special way, which it does by default.

### DIFF
--- a/lib/gollum-lib/markups.rb
+++ b/lib/gollum-lib/markups.rb
@@ -58,7 +58,7 @@ module Gollum
     if gem_exists?('pandoc-ruby')
       GitHub::Markup::Markdown::MARKDOWN_GEMS.delete('kramdown')
       GitHub::Markup::Markdown::MARKDOWN_GEMS['pandoc-ruby'] = proc { |content|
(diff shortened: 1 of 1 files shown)


heavywatal:
This change might be surprising for pandoc-ruby users not using MathJax. Can you make it follow `--mathjax` flag? i.e., your code if `--mathjax=true`, the original code if `--mathjax=false`.

dometto:
I agree that @heavywatal's suggestion would be ideal, but I'm afraid we may not have access to the mathjax setting at this point in the code. That is, I don't believe we have access to an instance of the `Wiki` class here. I would suggest that the change for non-mathjax using pandoc-users is acceptable -- I think they can always use backslashes to escape anything that will now be rendered as math?

heavywatal:
> I would suggest that the change for non-mathjax using pandoc-users is acceptable

Fair enough.

PR #385, opened 2020-09-02
Don't expand filesystem paths

toupeira:
Paths starting with a `~` tilde character were getting expanded by the
call to `File.expand_path` in `BlobEntry.normalize_dir`. This can cause
an exception when the tilde is followed by an invalid username, which
makes the whole wiki unusable.

This overrides `BlobEntry.normalize_dir` so it doesn't expand tildes
anymore, and extends the tests a bit.

There's another call to `File.expand_path` in `File.get_disk_reference`
which could be exploited similarly, this one was just removed because it
doesn't seem to be needed.

### DIFF
--- a/lib/gollum-lib/file.rb
+++ b/lib/gollum-lib/file.rb
@@ -171,7 +171,7 @@ module Gollum
       return false if @blob.is_symlink && !FS_SUPPORT_SYMLINKS
 
       # This will try to resolve symbolic links, as well
-      pathname = Pathname.new(::File.expand_path(::File.join(@wiki.repo.path, '..', @path)))
+      pathname = Pathname.new(::File.join(@wiki.repo.path, '..', @path))
       if pathname.symlink?
         source   = ::File.readlink(pathname.to_path)
         realpath = ::File.join(::File.dirname(pathname.to_path), source)
--- a/lib/gollum-lib/file.rb
+++ b/lib/gollum-lib/file.rb
@@ -171,7 +171,7 @@ module Gollum
       return false if @blob.is_symlink && !FS_SUPPORT_SYMLINKS
 
       # This will try to resolve symbolic links, as well
-      pathname = Pathname.new(::File.join(@wiki.repo.path, '..', @path))
+      pathname = Pathname.new(::File.join(@wiki.repo.path, '..', BlobEntry.normalize_dir(@path)))
       if pathname.symlink?
         source   = ::File.readlink(pathname.to_path)
         realpath = ::File.join(::File.dirname(pathname.to_path), source)
--- a/lib/gollum-lib/blob_entry.rb
+++ b/lib/gollum-lib/blob_entry.rb
@@ -80,22 +80,21 @@ module Gollum
     def self.normalize_dir(dir)
       return unless dir
 
-      # Return empty string for paths that point to the toplevel
-      # ('.', '/', '..', 'C:/' etc.)
-      return '' if dir =~ %r{\A([a-z]:\/)?[\./]*\z}i
+      dir = dir.dup
 
-      # Normalize the path:
-      # - Add exactly one leading slash
-      # - Remove trailing slashes
-      # - Remove repeated slashes
-      dir.sub(%r{
-        \A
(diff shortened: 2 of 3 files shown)


toupeira:
@dometto here's the promised PR, can you take a look? :grinning:

lib/gollum-lib/file.rb
```diff
@@ -171,7 +171,7 @@ def get_disk_reference
       return false if @blob.is_symlink && !FS_SUPPORT_SYMLINKS
 
       # This will try to resolve symbolic links, as well
-      pathname = Pathname.new(::File.expand_path(::File.join(@wiki.repo.path, '..', @path)))
```
dometto, line 174: Looking at this, my guess is that the `expand_path` call here is supposed to ensure that file paths are always under the repo root, in order to counter potential directory traversal exploits: passing in something like `../../../etc/password`:

```ruby
2.6.3 :002 > File.join("/path/to/repo", "../../../etc/passwd")
 => "/path/to/repo/../../../etc/passwd" 
```

Would it be an idea to:

1. Also remove `../` from paths in `BlobEntry.normalize_dir`.
2. Call `BlobEntry.normalize_dir(@path)` here to make sure such situations cannot arise

?

dometto (commented):
Apologies for not having given feedback on this earlier. As you may know the maintainers of this project are all volunteers, so between work and holidays thing have been a bit slow over the summer.

dometto:
Also: many thanks for spotting the problem and proposes this fix, @toupeira!

One additional question just to make sure: isn't it an option to just catch and rescue the exception thrown when passing in a 'wrong' user name? That would probably be faster. But I guess the downside it will actually allow expanding paths into users' directories?

lib/gollum-lib/file.rb
dometto: On second thought, that cannot be right, as its the result of the `File.join` that is being expanded. Nevertheless: isn't this something for which we should guard?

dometto: Perhaps such normalization/sanitization should be performed anyway at https://github.com/gollum/gollum-lib/blob/512b9667e37f35f2661cef1afce78080f715e439/lib/gollum-lib/file.rb#L77

@bartkamphorst if you have a moment, please let us know what you think.

toupeira: @dometto ah that makes sense! I changed `BlobEntry.normalize_dir` now to also strip `.` and `..`, and realized I was previously also returning `''` for a path like `...` which is still valid :wink: So I changed it around a bit to catch those cases, and added the call here in `get_disk_reference`.

I had some unrelated tests failing locally, so let's see if it passes the CI checks.

toupeira:
> Apologies for not having given feedback on this earlier. As you may know the maintainers of this project are all volunteers, so between work and holidays thing have been a bit slow over the summer.

@dometto no worries at all! :+1: 

> One additional question just to make sure: isn't it an option to just catch and rescue the exception thrown when passing in a 'wrong' user name? That would probably be faster. But I guess the downside it will actually allow expanding paths into users' directories?

That would work too, the expansion itself should be unproblematic if we raise an exception immediately afterwards. The main issue is that you then couldn't actually have a directory starting with a tilde, which some users might want :grinning:

lib/gollum-lib/file.rb
toupeira: @dometto oh I somehow only saw your first comment in this thread, let's see what @bartkamphorst thinks.

bartkamphorst: Sorry for wait. I like the proposed solution, and prefer it over rescuing the exception and losing the option of having directories start with a tilde. 👍 Thanks @toupeira for your work on this!

dometto:
@toupeira sorry for the wait, I didn't see the additions you made to the PR. Thanks for guarding against traversal in the new `normalize_dir`, and for including regression tests -- and generally for submitting this fix!

toupeira:
@dometto @bartkamphorst sorry this went under in my inbox :see_no_evil: Thanks to you both too!

PR #387, opened 2020-12-01
Consider wiki global metadata in PandocBib filter. Resolves https://github.com/gollum/gollum/issues/1627

dometto:


dometto:
Resolves https://github.com/gollum/gollum/issues/1627

PR #388, opened 2020-12-10
Update gems

dometto:


gemspec.rb
```diff
@@ -31,34 +31,31 @@ def specification(version, default_adapter, platform = nil)
...
     # required by pry
     s.add_development_dependency 'rb-readline', '~> 0.5.1'
     # updating minitest-reporters requires a new minitest which fails with gollum's tests.
-    s.add_development_dependency 'test-unit', '~> 3.1.5'
+    s.add_development_dependency 'test-unit', '~> 3.3'
     s.add_development_dependency 'minitest-reporters', '~> 1.4'
     s.add_development_dependency 'nokogiri-diff', '~> 0.2.0'
-    # required by guard
-    s.add_development_dependency 'guard', '~> 2.8.2'
-    s.add_development_dependency 'guard-minitest', '~> 2.3.2'
```
dometto, line 57: Looks like guard installs these dependencies itself now, so no need to require them ourselves.

PR #389, opened 2020-12-18
Fix broken TOC when some headers are ommitted. Resolves https://github.com/gollum/gollum/issues/1645

ViChyavIn:
When some of the headers are ommitted, TOC is not displayed properly. This commit fixes the issue.

image


lib/gollum-lib/filter/toc.rb
```diff
@@ -59,6 +68,17 @@ def process(data)
 
   private
 
+  # Generates header in format "h<level>"
+  def make_header_tag(level)
+    raise "Header should be from 1 to 6" unless level.between?(1, 6)
```
dometto, line 73: Shouldn't this exception be handled somewhere?

lib/gollum-lib/filter/toc.rb
```diff
@@ -59,6 +68,17 @@ def process(data)
...
+    "h#{level}"
+  end
+
+  def find_offset(level)
+    tags_before = @missing_headers.select { |number| number < level }
```
dometto, line 78: Not sure yet if I completely understand how this code fixes the bug. What if we have the following situation:

```
h2
h1
...
```

There isn't an h1 before the h2, so presumably this would trigger the bug? But as far as I can see `h1` would not be in `@missing_headers`, because it *is* present in the document. What am I missing?

dometto (commented):
Thanks again! Some questions in the review comments. As always, but especially with this PR, a test case would be very welcome. History has shown that without tests, it's very easy to re-introduce bugs in this rather complex code when refactoring it later. :)

lib/gollum-lib/filter/toc.rb
ViChyavIn: Yes, the PR doesn't handle this situation. I have a wiki and sometimes people prefer using, let's say, h2 headers instead of h1, so they omit h1 headers throughout all the page. `@missing_headers` contain those headers that are missing at all (if it occures at least once, it will reintroduce the bug).

I can try to rewrite my PR so it fixes your case as well, is it necessary?

ViChyavIn:
Rewrote the fix in 4df644d so now it should handle all the cases, including:
```
h2
h1
...
```
and added some tests.

dometto:
Great work! Many thanks!

PR #397, opened 2021-02-13
Proof of concept: Rmarkdown support

dometto:
Resolves https://github.com/gollum/gollum/issues/1553

I managed to implement support for RMarkdown relatively painlessly by modifying the code filter. Obviously this is only a proof of concept. The biggest problem is that this implementation relies on the RinRuby gem which seems to be no longer maintained. I can't get the `master` version of that gem to work (the app hangs on startup when I `require 'rinruby'`), but the older released gem [has a bug](https://github.com/clbustos/rinruby/issues/7) (fixed on `master`) which means you can't `Ctrrl-D` to interrupt the R process. :( Nevertheless, if there are different R-for-Ruby implementations, we would only have to modify the `render_r` method in the code filter.

Anyway, the end result on gollum is pretty cool:

Screenshot 2021-02-13 at 20 52 20

For this source:

```
```{r first_r_chunk}
vec <- c(1, 2, 3)
print(vec)
```

Tset

```{r second_r_chunk, echo = FALSE}
vec2 <- c(10, 20, 30)
vec3 <- vec * vec2
print(vec3)     
```

Test

```{r echo = FALSE}
library(ggplot2)
data(mpg, package="ggplot2")

mpg_select <- mpg[mpg$manufacturer %in% c("audi", "ford", "honda", "hyundai"), ]

# Scatterplot
theme_set(theme_bw())  # pre-set the bw theme.
g <- ggplot(mpg_select, aes(displ, cty)) + 
  labs(subtitle="mpg: Displacement vs City Mileage",
       title="Bubble chart")

g + geom_jitter(aes(col=manufacturer, size=hwy)) + 
  geom_smooth(aes(col=manufacturer), method="lm", se=F)
```% 
```

### DIFF
--- a/lib/gollum-lib/filter/code.rb
+++ b/lib/gollum-lib/filter/code.rb
@@ -3,9 +3,7 @@
 #
 # Render a block of code using the Rouge syntax-highlighter.
 
-R = :nil
 require 'rinruby'
-RInterface = RinRuby.new
 
 class Gollum::Filter::Code < Gollum::Filter
   def extract(data)
@@ -159,9 +157,9 @@ class Gollum::Filter::Code < Gollum::Filter
   end
   
   def render_r(content)
-    RInterface.eval 'require(knitr)'
-    RInterface.assign "content", content
-    RInterface.eval "knitr::render_markdown(strict = TRUE)"
-    RInterface.pull "(knitr::knit2html(text = content, fragment.only = TRUE))"
+    R.eval 'require(knitr)'
+    R.assign "content", content
+    R.eval "knitr::render_markdown(strict = TRUE)"
+    R.pull "(knitr::knit2html(text = content, fragment.only = TRUE))"
   end
 end
--- a/lib/gollum-lib/filter/code.rb
+++ b/lib/gollum-lib/filter/code.rb
@@ -1,8 +1,12 @@
 # ~*~ encoding: utf-8 ~*~
-
 # Code
 #
 # Render a block of code using the Rouge syntax-highlighter.
+
+R = :nil
+require 'rinruby'
+RInterface = RinRuby.new
+
 class Gollum::Filter::Code < Gollum::Filter
   def extract(data)
     case @markup.format
@@ -33,9 +37,16 @@ class Gollum::Filter::Code < Gollum::Filter
       end  
     end
     
-
     data.gsub!(/^([ ]{0,3})``` ?([^\r\n]+)?\r?\n(.+?)\r?\n[ ]{0,3}```[ \t]*\r?$/m) do
-      "#{Regexp.last_match[1]}#{cache_codeblock(Regexp.last_match[2].to_s.strip, Regexp.last_match[3], Regexp.last_match[1])}" # print the SHA1 ID with the proper indentation
+      m_indent = Regexp.last_match[1]
+      m_extra   = Regexp.last_match[2]
+      m_code   = Regexp.last_match[3]
+      if m_extra =~ /^\{(\s+)?r(\s.*)?\}/
+        replacement = cache_rblock(m_extra.to_s.strip, m_code, m_indent)
+      else
+        replacement = cache_codeblock(m_extra.to_s.strip, m_code, m_indent)
+      end
+      "#{Regexp.last_match[1]}#{replacement}" # print the SHA1 ID with the proper indentation
     end
     data
   end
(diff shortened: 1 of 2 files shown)


dometto:
Fun fact: running it on JRuby with the `master` branch of https://github.com/clbustos/rinruby, everything works nicely!

Running with this in gollum's `Gemfile`:

```ruby
gem 'rinruby', :github => 'clbustos/rinruby'
gem 'gollum-lib', :path => '/Users/dawa/Code/gollum_project/gollum-lib'
```

dometto:
The difference between JRuby and MRI turns out to be that on JRuby, `rinruby`'s so-called "interactive mode" was disabled by default. Disabling it explicitly fixes the problems on MRI, too.

bartkamphorst:
I wonder if we need `rinruby` or whether we would be better off rolling our own?

lib/gollum-lib/filter/code.rb
```diff
@@ -126,14 +136,34 @@ def remove_leading_space(code, regex)
...
-    id = "#{open_pattern}#{Digest::SHA1.hexdigest("#{language}.#{code}")}#{close_pattern}"
+    id = generate_placeholder_id(language, code)
     cached = @markup.check_cache(:code, id)
     @map[id] = cached ?
       { :output => cached } :
       { :lang => language, :code => code, :indent => indent }
     id
   end
+  
+  def render_r(content)
```
bartkamphorst, line 160: Can you explain how the resulting figure is dealt with? Is the image stored somewhere on disk? Or is it `base64` encoded and placed directly in the html fragment?

dometto: The latter! `knitr` ensures every fragment is rendered directly "on screen".

dometto:
> I wonder if we need rinruby or whether we would be better off rolling our own?

I've wondered about this too. Three points:

* `rinruby` seems to work fine out of the box when you just run it in `irb` for instance. It's the combination with sinatra that seems to cause trouble unless interactive mode is disabled.
* with interactive mode disabled, it seems to work fine with sinatra as well.
* there are some [complicated bits](https://github.com/clbustos/rinruby/blob/master/lib/rinruby.rb#L450-L648) that I wouldn't look forward to writing myself (or would I?).

The maintainability of the project does not look great, but maybe it's better than nothing for MRI? [On Java](https://www.baeldung.com/java-r-integration), there seem to be some other options, and on Truffle there would be https://github.com/rbotafogo/galaaz.

Another way to go would be using a TCP client to interact with [RServe](https://www.baeldung.com/java-r-integration#rserve). There is an old [ruby client](https://github.com/clbustos/Rserve-Ruby-client) for that by the same maintainer as `rinruby`. But it would require people to have installed and run RServe (not just R).

dometto:
Some more notes:

* The current approach requires the `knitr` (and `gnuplot2`) R packages to be installed by the user, but I think there's no real substitute for that (we don't really want to go installing packages on people's machines automatically).
* We should test whether the speed of this implementation is acceptable for some more serious computations.
* Currently, we're rendering `{r`} code blocks on pages in every format, not just `.rmd`. But maybe that's actually a cool feature? ROrg-Mode with gollum!

dometto:
I should also clarify that when I say it works fine without interactive mode, that's on the `master` branch of `rinruby`. On the latest released version of the gem, everything "works" but there's still the problem of it being impossible to kill the app via ctrl-c. So there's a problem of maintainability anyway.

lib/gollum-lib/filter/code.rb
bartkamphorst: That's very neat! I must have missed that option when researching this before.

bartkamphorst:
> The current approach requires the knitr (and gnuplot2) R packages to be installed by the user, but I think there's no real substitute for that (we don't really want to go installing packages on people's machines automatically).

Agreed, but we could try to render a useful message instead of either breaking or not rendering.

dometto:
Conceptually, for MRI, I actually like the approach taken by this [old gem](https://github.com/adhitia/rsruby) which leverages `libR` through a C-extension to connect with an R instance -- so no need to handle sockets and threads in ruby. I don't know if a modern approach to that using `ffi` would work on JRuby.

dometto:
Also: we should probably disable R execution by default for security purposes. Users who want to use RMarkdown on a public instance of gollum should probably use something like https://github.com/jeroen/RAppArmor.

dometto:
Here's a way of capturing the output of `R --slave` (listen on standard in) that works on JRuby as well as MRI: https://gist.github.com/dometto/4b0bb1e43ef96f81d30c3d018062f64d

dometto:
Different implementation coming

PR #400, opened 2021-05-03
implement ranked search

jklippel:
In a project using gollum for documentation it came up that the search was not as helpful as it could be.
One has to search for "relevant" results in the result list.

The request came up to improve the search by implementing a rank.

A rank was implemented based on four values:

- Does the main heading match?
  - makes up to 50% of the ranks total 100 Points
  - one main heading match is enough to get the full 50%

- Does another heading match?
  - makes up to 25% of the rank
  - two heading matches are enough to get the full 25%

- Does the link description match?
  - makes up to 15% of the rank
  - two link matches are enough to get the full 15%

- Does the text match?
  - makes up to 10% of the rank
  - five matches are enough to get the full 10%

The rank is implemented by categorizing the search results into the four categories using counts (method count_matches).
The rank is then calculated by a call to the method weigh_rank.

The resulting rank is returned by search as a new member in the returned results hash (named :rank).

It can be included into the search output by adapting gollum. A separate pull-request will be made in the gollum repository.

The implementation does not break the current implementation only adds the :rank field to the search result.

### DIFF
--- a/lib/gollum-lib/wiki.rb
+++ b/lib/gollum-lib/wiki.rb
@@ -499,19 +499,78 @@ module Gollum
         result[:name] = extract_page_file_dir(name)
         result[:filename_count] = result[:name].scan(/#{search_terms_regex}/i).size
         result[:context] = []
+        result[:match_counts] = [0, 0, 0, 0]
         if data
           begin
             data.scan(query) do |match|
               result[:context] << match.first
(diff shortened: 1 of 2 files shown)


jklippel:
gollum pull-request: https://github.com/gollum/gollum/pull/1717

PR #402, opened 2021-07-09
sanitize macros

yy0931:
@dometto This is a PR which fixes the XSS vulnerability I reported to you by email just now.

I think using sanitize(data) as you said is a good idea because custom macros can also be escaped, but CGI::escapeHTML is also necessary to avoid outputting safe but broken HTML, so I used both.

### DIFF
--- a/lib/gollum-lib/filter/macro.rb
+++ b/lib/gollum-lib/filter/macro.rb
@@ -55,6 +55,6 @@ class Gollum::Filter::Macro < Gollum::Filter
       end
     end
 
-    data
+    sanitize(data)
   end
 end
--- a/lib/gollum-lib/macro/all_pages.rb
+++ b/lib/gollum-lib/macro/all_pages.rb
@@ -3,7 +3,7 @@ module Gollum
     class AllPages < Gollum::Macro
       def render
         if @wiki.pages.size > 0
-          '<ul id="pages">' + @wiki.pages.map { |p| "<li>#{p.name}</li>" }.join + '</ul>'
+          '<ul id="pages">' + @wiki.pages.map { |p| "<li>#{CGI::escapeHTML(p.name)}</li>" }.join + '</ul>'
         end
       end
     end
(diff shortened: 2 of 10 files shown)


dometto:
@yy0931 I think the problem with the JRuby tests is that sanitize is adding `<p>` tags around the video tags, not the order of the attributes. We can safely ignore this for now.

dometto:
@benjaminwil @bartkamphorst for context, the problem that @yy0931 discovered that we are executing the `Sanitize` filter *before* we are rendering Macros. The fix in this PR is perhaps not ideal, as it means we're sanitizing the entire document twice: once in the `Sanitize` filter and once in the `Macro` filter. Perhaps it's better to move the `Sanitize` filter to the first position in the filter chain, so that sanitization is the last step. That depends on two questions:

1. Would that cause any security issues? (I think not)
2. Would that break any other filters? (I'm not sure)

So I'd like to merge this fix so we can do a quick update of the gem, so the vulnerability is at least gone. Then we can see about what the best approach would be -- feedback welcome!

Unfortunately, the JRuby tests are now failing:

```
   </<video (.*) (.*) src="\/Uploads\/foo.mp4"(.*)> (.*)<\/video>/> was expected to be =~
  <"<p><video height=\"100%\" src=\"/Uploads/foo.mp4\" width=\"100%\"> HTML5 video is not supported on this Browser.</video></p>\n">.
```

Looks like sanitize on JRuby is adding `<p>` tags around the video tags. I think we can live with this, so I'll just merge as is, and then we can adapt the tests later.

The last two times I tried to get a CVE assigned to an exploit in gollum I never received a reply, so I'm not really keen to go down that route again. :-(

Many thanks to @yy0931!

yy0931:
I have no idea why but the `<p>` tag seems to only appears in the test error messages.

dometto:
Ah, JRuby tests fixed! Thanks again @yy0931!

I'll merge this now and do a Gem release. Suggestions on how to advertise the need to update welcome. I'll add a line to the History file anyway.

benjaminwil:
I'd be interested to find out if moving the sanitization filter to the end of the filter chain breaks any of the other filters. I can see that being possible.

Thanks for pinging me on this ticket. Very interesting. 😄

PR #409, opened 2022-01-14
Improve code filter performance

dometto:
I thought I saw an opportunity for improving the performance of the code filter. Instead of calling `data.gsub!` for every `id` in `@map`, we can call `data.gsub!` only once, and have it match all substitution patterns.

This refactor passes the tests, but they don't run noticeably faster. @bartkamphorst can you remind me how to do a quick performance test?


bartkamphorst:
Do you mean `Benchmark`? 
```ruby
Benchmark.bm do |x|
  x.report { for i in 1..n; a = "1"; end }
  x.report { n.times do   ; a = "1"; end }
  x.report { 1.upto(n) do ; a = "1"; end }
end
```

PR #410, opened 2022-01-14
Implement language handlers in Code Filter

dometto:
This PR implements customizable behavior for handling code blocks in specific languages in the `Code` filter. Usually, we want to simply highlight a code block. But in some cases, we (or the user) might want to skip highlighting and instead process a code block in a different way. Two possible use cases:

* Implementing support for [Mermaid](https://github.com/gollum/gollum/issues/1473)
* Implementing support for [RMarkdown](https://github.com/gollum/gollum/issues/1553)

In both cases, we want to preserve the code in the block so that it can be processed in a different way later.

With the current implementation, it's trivial to define custom behavior for new languages: just add a pattern and a `Proc` object to `Gollum::Filter::Code`'s `@@language_handlers` hash.

We could also choose to keep that hash empty in `gollum-lib`, and instead define the handlers in `gollum` (if we don't want to have more frontend-specific HTML in `gollum-lib`). To illustrate the functionality, I've just defined two handlers for Mermaid and RMarkdown here.

Note: the idea behind this is that we don't really want to have a separate filter for Mermaid and RMarkdown (and whatever else), since that would require going through the entire document to find/replace codeblocks in multiple filters, which would be inefficient.

### DIFF
--- a/test/filter/test_code.rb
+++ b/test/filter/test_code.rb
@@ -10,6 +10,18 @@ context "Gollum::Filter::Code" do
     @filter.process(@filter.extract(content))
   end
   
+end
+
+context "Gollum::Filter::Code with language handler" do
+  setup do
+    @filter = Gollum::Filter::Code.new(Gollum::Markup.new(mock_page))
+    Gollum::Filter::Code.language_handlers[/mermaid/] = Proc.new { |lang, code| "<div class=\"mermaid\">\n#{code}\n</div>" }
+  end
+
+  def filter(content)
+    @filter.process(@filter.extract(content))
+  end
+  
   test 'mermaid language handler' do
     markup = <<~EOF
     # Some markup
@@ -29,5 +41,8 @@ context "Gollum::Filter::Code" do
     assert filter(markup), result
   end
   
+  teardown do
+    Gollum::Filter::Code.language_handlers = {}
+  end
   
 end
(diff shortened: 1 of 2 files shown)


test/filter/test_code.rb
```diff
@@ -0,0 +1,33 @@
...
+    sequenceDiagram
+      Alice->>John: Hello John, how are you?
+      John-->>Alice: Great!
+      Alice-)John: See you later!
+    ```
+    
+    Foo
+    EOF
+    
+    result = "# Some markup\n\n<div class=\"mermaid\">\nsequenceDiagram\n  Alice-&gt;&gt;John: Hello John, how are you?\n  John--&gt;&gt;Alice: Great!\n  Alice-)John: See you later!\n</div>\n\nFoo\n"
```
dometto, line 27: Note that the ouput is HTML escaped, but fortunately Mermaid correctly renders the escaped markup (see [here](https://github.com/gollum/gollum/issues/1473#issuecomment-1013160183))!

bartkamphorst: 👍

bartkamphorst:
I like this approach. I would prefer it if `@@language_handlers` were set in gollum.

andrewlytle:
Any update on this? Would love to have this functionality in Gollum

dometto:
Refactored this a bit to let the language handlers be easily configurable from gollum. It now uses effectively the same approach as our `Markup` class uses to 'register' formats and extensions (just call the `language_handlers` class method to access and modify the `language_handlers` Hash).

PR #415, opened 2022-02-16
Conditionally render "editable" heading classes

benjaminwil:
In the `gollum` gem, wikis are initialized with a setting `allow_editing`. So far, this gem is unaware of that setting and does nothing with it.

However, this gem does consider some headings in a document "editable" and other headings not editable. It specifies a CSS class, "editable" on any heading it considers editable.

This commit gives `gollum-lib` more context about whether a heading should be editable or not based on the settings of the current `Wiki`. The default value is `true`, but now we can set the `allow_editing` setting to specify whether a heading in a document should be rendered with the "editable" class or not.

### DIFF
--- a/lib/gollum-lib/filter/render.rb
+++ b/lib/gollum-lib/filter/render.rb
@@ -19,8 +19,11 @@ class Gollum::Filter::Render < Gollum::Filter
   end
 
   def process(data)
-    data = add_editable_header_class(data)
-    data
+    if @markup.wiki.allow_editing
+      add_editable_header_class(data)
+    else
+      data
+    end
   end
 
   private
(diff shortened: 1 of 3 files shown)


benjaminwil:
I've broken a seemingly unrelated spec, and I haven't figured out why yet.

dometto:
This seems like a very reasonable approach! Thanks @benjaminwil

lib/gollum-lib/filter/render.rb
```diff
@@ -30,6 +30,8 @@ def add_editable_header_class(data)
     doc.css('h1,h2,h3,h4,h5,h6').each_with_index do |header, i|
       next if header.content.empty?
       next if header.inner_html.match(PLACEHOLDER_PATTERN)
+      next unless @markup.wiki.allow_editing
```
dometto, line 33: We could also skip the entire `add_editable_header_class` call if `@markup.wiki.allow_editing` is `false`

dometto:
I'm also not sure what broke the spec...

lib/gollum-lib/filter/render.rb
benjaminwil: Good suggestion. Thank you!

benjaminwil:
@dometto Apparently the whitespace that my editor removed was actually important. 😅 

I've fixed the failing spec. If this looks good to you I can merge tomorrow.

dometto (approved):
:fireworks:

PR #416, opened 2022-03-24
Add Ruby 3.1 to the CI matrix

mishina2228:
Same change as in gollum/gollum#1812. 😄 

---

This PR adds Ruby 3.1 to the CI matrix.

Using 3.0 without quotes, ruby/setup-ruby will use 3.1 instead of 3.0.
To avoid similar parsing issues, all version numbers are enclosed in quotes.


dometto:
Great thanks @mishina2228!

PR #417, opened 2022-03-25
Fix deprecation warnings

mishina2228:
This PR suppresses deprecation warnings at test running.
It includes the following:
- Pass a Document as the second parameter to `Nokogiri::XML::Node.new`.
This will suppress the following warning:
  ```
  warning: Passing a Node as the second parameter to Node.new is deprecated. Please pass a Document instead, or prefer an 
  alternative constructor like Node#add_child. This will become an error in a future release of Nokogiri.
  ```
  Please see the following for more details:
  https://github.com/sparklemotion/nokogiri/issues/975
- Require 'mocha/test_unit' instead of 'mocha/setup'
This will suppress the following deprecation warning:
  ```
  Mocha deprecation warning at /home/runner/work/gollum-lib/gollum-lib/test/helper.rb:9:in `require': Require 'mocha/test_unit', 'mocha/minitest' or 'mocha/api' instead of 'mocha/setup'.
  ```

### DIFF
--- a/lib/gollum-lib/filter/toc.rb
+++ b/lib/gollum-lib/filter/toc.rb
@@ -125,7 +125,7 @@ class Gollum::Filter::TOC < Gollum::Filter
   # Creates an anchor element with the given name and adds it before
   # the given header element.
   def add_anchor_to_header(header, name)
-    a = Nokogiri::XML::Node.new('a', @doc)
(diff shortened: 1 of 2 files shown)


benjaminwil:
Thank you!

PR #418, opened 2022-04-05
Remove gemspec fields that should not be set

mishina2228:
Similar changes as in gollum/gollum#1811.

RubyGems says not to set the value to [`date`](https://github.com/rubygems/rubygems/blob/fe96fb6e2ac5a8b6df5e852470d11fa854301eca/lib/rubygems/specification.rb#L1677-L1682), [`specification_version`](https://github.com/rubygems/rubygems/blob/fe96fb6e2ac5a8b6df5e852470d11fa854301eca/lib/rubygems/specification.rb#L750-L755) and [`rubygems_version`](https://github.com/rubygems/rubygems/blob/fe96fb6e2ac5a8b6df5e852470d11fa854301eca/lib/rubygems/specification.rb#L536-L541).
Also, `rubyforge_project` is deprecated.

This PR removes those fields from gemspec.
And does the following:
- Fix gemspec task not to set date and rubyforge_project field.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,6 +1,5 @@
 require 'rubygems'
 require 'rake'
-require 'date'
 
 #############################################################################
 #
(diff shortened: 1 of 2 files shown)


dometto:
thanks @mishina2228!

PR #420, opened 2022-05-24
Update octicons

dometto:
Looked at the breaking changes for `octicons` since v.12:

```
        THIS IS A BREAKING CHANGE and will require re-linking all the file-directory icon references to file-directory-fill
	•	Rename 16px select-single icon to single-select #665
	•	Rename duplicate icon to copy #643
	•	Rename clippy icon to paste #643
	•	Remove octoface #611
	•	Rename git-fork-24 to repo-forked-24 #593
	•	Remove 24px insights icon #574
	•	Remove 24px copy icon #586
```

None should affect gollum-lib.

### DIFF
--- a/gemspec.rb
+++ b/gemspec.rb
@@ -26,7 +26,7 @@ def specification(version, default_adapter, platform = nil)
     s.add_dependency 'loofah', '~> 2.3'
     s.add_dependency 'github-markup', '~> 4.0'
     s.add_dependency 'gemojione', '~> 4.1'
-    s.add_dependency 'octicons', '~> 12.0'
+    s.add_dependency 'octicons', '~> 17.0'
     s.add_dependency 'twitter-text', '1.14.7'
 
     s.add_development_dependency 'org-ruby', '~> 0.9.9'


dometto:
At the moment, updating `octicons` in this project will break gollum, because gollum still depends on an older version of that gem. But updating `octicons` in gollum is also impossible as long as we haven't updated it in `gollum-lib`. So unless we cut a new major-level release, it looks as if we're stuck in a dependency loop. I am not sure how best to address this. One option would be to turn it into a development dependency, but that does mean parts of this lib will not be self-sufficient anymore. Any thoughts @bartkamphorst?

bartkamphorst:
I actually think changing the icon set can be so disruptive that it can warrant a major release. Not sure whether the changes to the actual icons are that big though, but I don't mind waiting with the update of `octicons` until gollum `6.x`.

dometto:
We can of course wait for 6.0 in this case, but the problem will persist if we ever do want to update octicons  before a major release, so I'm just wondering what the best way to handle the dependency is. Should we perhaps just remove the dependency from `gollum`? That feels weird because it's such a "front-endy" dependency, but it would resolve the dependency loop.

dometto:
ping @bartkamphorst, I forgot about this issue but would be great to hear your thoughts on the dependency loop.

dometto:
To clarify, I have updated octicons in gollum-lib in this PR, but the question of the dependency loops still persists. If there's a breaking change to `octicons`, we have to update the dependency in, and release new gems for, both `gollum` and `gollum-lib`.

PR #422, opened 2022-05-24
Try updating dependencies

dometto:
Just seeing what will break on CI :)


dometto:
`twitter-text` latest version errors with `ERROR: could not find idn library!`

dometto:
`RedCloth > 4.2.9` fails on JRuby (no longer supported). Since this is just a development dependency, that's not such a big deal.

PR #424, opened 2022-08-31
Implement default branch detection, move away from master as default.

dometto:
Resolves https://github.com/gollum/gollum/issues/1813

Requires these changes to the adapter: https://github.com/gollum/rugged_adapter/pull/55 (RJGit adapter support forthcoming)

This is the easiest way I could see to support both `master` and `main` as default branches. The default ref will be determined as follows, in order:

1. is there an explicit branch set by the user (`--ref`)?
2. If not, try the branch names in `Wiki.default_refs` in order. If any of them exist, default to it.
3. If not, see if there is a global git default branch name set and use it.
4. If not, use `main`

NB: if step 2 fails, we still want to default to something, as the `gollum-lib` API supports working with an empty repo (gollum can make the first commit, but then needs to know on what branch).

Would this be an acceptable solution? A simpler alternative would be to just drop support for `master` altogether, but I think a middle way might be good.

Questions:

* What do we want to have in `Wiki.default_refs`? Presently I went for `['main', 'master']`. Anything else?
* Do we want to print the deprecation message for `master` at all? Or print it only outside of tests?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,4 +1,9 @@
 source 'https://rubygems.org'
 gemspec :name => 'gollum-lib'
 gem 'irb'
-gem 'gollum-rugged_adapter', :git => 'https://github.com/gollum/rugged_adapter/', :branch => 'find_branches'
+
+if RUBY_PLATFORM == 'java'
+  gem 'gollum-rjgit_adapter', git: 'https://github.com/dometto/gollum-lib_rjgit_adapter/', branch: 'find_branch'
+else
+  gem 'gollum-rugged_adapter', git: 'https://github.com/gollum/rugged_adapter/', branch: 'find_branches'
+end
--- a/Gemfile
+++ b/Gemfile
@@ -1,3 +1,4 @@
 source 'https://rubygems.org'
 gemspec :name => 'gollum-lib'
 gem 'irb'
+gem 'gollum-rugged_adapter', :git => 'https://github.com/gollum/rugged_adapter/', :branch => 'find_branches'
--- a/Gemfile
+++ b/Gemfile
@@ -1,3 +1,3 @@
 source 'https://rubygems.org'
-
 gemspec :name => 'gollum-lib'
+gem 'irb'
--- a/Gemfile
+++ b/Gemfile
@@ -5,6 +5,4 @@ gem 'irb'
 # Chaanges to this branch require corresponding changes in the adapters, see https://github.com/gollum/gollum-lib/pull/424
 if RUBY_PLATFORM == 'java'
   gem 'gollum-rjgit_adapter', git: 'https://github.com/dometto/gollum-lib_rjgit_adapter/', branch: 'find_branch'
-else
-  gem 'gollum-rugged_adapter', git: 'https://github.com/gollum/rugged_adapter/', branch: 'find_branches'
 end
--- a/Gemfile
+++ b/Gemfile
@@ -2,3 +2,8 @@ source 'https://rubygems.org'
 gemspec :name => 'gollum-lib'
 gem 'irb'
 
+if RUBY_PLATFORM == 'java' then
+  group :development do
+    gem 'activesupport', '~> 6.0'
+  end
+end
--- a/Gemfile
+++ b/Gemfile
@@ -2,6 +2,7 @@ source 'https://rubygems.org'
 gemspec :name => 'gollum-lib'
 gem 'irb'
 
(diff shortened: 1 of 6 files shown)


dometto:
The test failures are due to the following:

1. The global git `init.defaultBranch` is not set on CI
2. git still appears to set `HEAD` to `master` instead of `main` by default (hardcoded)
3. Our tests setup test repos using `git` on the command line, which thus sets HEAD to `master`, but doesn't actually create a `master` branch (i.e. there are no commits).
4. Since there are no commits on `master`, gollum doesn't detect a `master` branch and defaults to `main` -> the wiki's ref and the repo's HEAD don't match.

There are two possible solutions:

a. Change the order of default refs to `['master', 'main']` instead of the other way around
b. Set `init.defaultBranch` on CI

Perhaps a. is more safe, since users might run into the same problem when creating a new wiki with `git init`?

dometto:
> There are two possible solutions:
> a. Keep the final back ref set to master instead of main (but still support main)
> b. Set init.defaultBranch on CI

> Perhaps a. is more safe, since users might run into the same problem when creating a new wiki with git init?

I implemented it this way for now, tests are now only failing on JRuby due to lack of changes to the RJGit adapter.

@bartkamphorst @benjaminwil would appreciate any thoughts!

Gemfile
```diff
@@ -1,3 +1,4 @@
 source 'https://rubygems.org'
 gemspec :name => 'gollum-lib'
 gem 'irb'
+gem 'gollum-rugged_adapter', :git => 'https://github.com/gollum/rugged_adapter/', :branch => 'find_branches'
```
benjaminwil: I wonder if it's worth leaving a code comment with a link to the issue inline? That way if more changes are made to this line or this file before `find_branches` is merged, the context needed to understand the issue won't be be three clicks away in a PR description.

Gemfile
```diff
@@ -1,4 +1,9 @@
 source 'https://rubygems.org'
 gemspec :name => 'gollum-lib'
 gem 'irb'
-gem 'gollum-rugged_adapter', :git => 'https://github.com/gollum/rugged_adapter/', :branch => 'find_branches'
+
+if RUBY_PLATFORM == 'java'
+  gem 'gollum-rjgit_adapter', git: 'https://github.com/dometto/gollum-lib_rjgit_adapter/', branch: 'find_branch'
+else
+  gem 'gollum-rugged_adapter', git: 'https://github.com/gollum/rugged_adapter/', branch: 'find_branches'
+end
```
benjaminwil, line 9: Similar comment re: inline comment here. Having the context readily available about the conditions that require this temporary change would be nice, in my opinion.

Rakefile
```diff
@@ -174,3 +181,61 @@ task :validate do
...
+  
+  temp = Tempfile.new
+  temp.puts("#{version_pattern} / #{date}\n#{body}\n")
+  temp.close
+  `cat #{history_file} >> #{temp.path}`
+  `cat #{temp.path} > #{history_file}`
+end
+
+desc 'Precompile assets'
+task :precompile do
```
benjaminwil, line 230: `gollum-lib` does not contain any assets and doesn't contain `sprockets` as a dependency. Does this need to be in the `gollum-lib` Rakefile?

benjaminwil (commented):
This makes sense to me.

From a git and PR history perspective: I am unsure how the first two commits relate to the feature described by this PR description. (I could just missing the connection.) If it's unnecessary for this change and could be better described by another PR I might argue we separate them out.

That said I think this is a good solution to the issue. the `--ref` is still configurable but we will handle 99 percent of users's default branches!

benjaminwil:
> Do we want to print the deprecation message for master at all? Or print it only outside of tests?

I don't think git itself prints deprecation messages for old repos to stop using `master`. I think it's reasonable to let people use `master` indefinitely without deprecation warnings.

dometto:
> From a git and PR history perspective: I am unsure how the first two commits relate to the feature described by this PR description. (I could just missing the connection.) If it's unnecessary for this change and could be better described by another PR I might argue we separate them out.

I had to rebase the 6.x branch on master to get rid of these commits, but that was worth doing -- thanks!

I'll do `2.0` releases of the adapters and make 6.x depend on them before I merge this.

Rakefile
dometto: This was copied from the `gollum` Rakefile in one of the commits that's already in `master`. I'll get rid of it in a separate PR!

Gemfile
dometto: Will do this for now, but as I said I'd prefer to remove this temp change altogether before merging this into 6.x!

PR #432, opened 2022-11-21
Allow defining custom rendering Procs for a format

dometto:
This change to the API allows users a bit more freedom to define custom formats. Currently, we are limited to formats that exist in `Github::Markup`. This PR allows us to bypass `Github::Markup` entirely when defining a format: instead, the `Render` filter will directly call a user-supplied block. 

As an example, with this change, I was able to get rudimentary support for [Fountain](https://github.com/gollum/gollum/issues/1715) working with the following snippet in `config.rb`:

```ruby
fountain_lua = '/path/to/.pandoc/filters/fountain.lua'
require 'pandoc-ruby'
Gollum::Markup.register(
    :fountain, "Fountain", :extensions => ['fountain'],
    :enabled => Gollum::MarkupRegisterUtils::gem_exists?('pandoc-ruby'),
    :render => proc {|content| ::PandocRuby.convert(content, f: fountain_lua, to: :html) }
)
```

But of course the `:render` `Proc` can contain anything, it's not limited to calling `PandocRuby`.

Feedback very welcome!

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,3 +1,8 @@
+# 5.2.1 / 2022-09-13
+
+* Fixed: 'controls' attribute in audio and video tags should not be sanitized. #430 (@dometto)
+
+
 # 5.2 / 2022-05-28
 
 * Conditionally render "editable" heading classes. Resolves https://github.com/gollum/gollum/issues/1785 (@benjaminwil)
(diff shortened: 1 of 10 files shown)


dometto:
Closing to properly rebase on 6.x.

PR #434, opened 2022-12-06
Preserve relative links for absent pages

dometto:
Resolves https://github.com/gollum/gollum/issues/1761

If a link to `Foo` from a page within a directory `subdir` is broken, we want the link to point to `subdir/Foo` instead of `/Foo`. (A broken absolute link to `/Foo` should point to `/Foo`, of course.)


dometto:
@bartkamphorst could you have a look to sanity check my code here?

dometto:
New approach thanks to live feedback from @bartkamphorst!

bartkamphorst (approved):
🎆

PR #437, opened 2023-01-13
Improve performance

dometto:
In response to https://github.com/gollum/gollum/issues/1940#issuecomment-1381623614 I did some profiling and saw that *a lot* of time was being spent in the `Gollum::File.find`, iterating over entries in the treemap. Basically, gollum currently iterates over a Hash representation of *all* the blobs and trees in the repo (at a given commit). In a large repo, this means looping a long time (plus all the inefficiencies derived from instantiating objects associated with each entry in the repo). I decided to hack together another approach that does not rely on a treemap, and it was relatively straightforward to do.

Here are some quick profiling results, loading a page from the `nlab-conten` wiki referred to by @felixwellen:

* Rugged:
  *master: 26653ms
  * this branch: 784ms
* RJGit
  * master: 27704ms
  * this branch: 2279ms

(I used https://github.com/MiniProfiler/rack-mini-profiler to profile)

On a smaller repo (our `lotr.git`), tested with rugged, the speed increase is not noticeable (but I don't think it's any slower either ;)).

Feedback welcome!

Todo:

- [x] Fix symlinks (requires work on the adapters)
- [x] Implement global lookup

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,10 +3,7 @@ gemspec :name => 'gollum-lib'
 gem 'irb'
 
 if RUBY_PLATFORM == 'java' then
-  gem 'gollum-rjgit_adapter', git: 'https://github.com/dometto/gollum-lib_rjgit_adapter', branch: 'tree_find_blob'
   group :development do
     gem 'activesupport', '~> 6.0'
   end
-else
-  gem 'gollum-rugged_adapter', git: 'https://github.com/gollum/rugged_adapter'
 end
--- a/Gemfile
+++ b/Gemfile
@@ -3,7 +3,7 @@ gemspec :name => 'gollum-lib'
 gem 'irb'
 
 if RUBY_PLATFORM == 'java' then
-  gem 'gollum-rugged_adapter', git: 'gollum-lib_rjgit_adapter', branch: 'tree_find_blob'
+  gem 'gollum-rjgit_adapter', git: 'https://github.com/dometto/gollum-lib_rjgit_adapter', branch: 'tree_find_blob'
   group :development do
     gem 'activesupport', '~> 6.0'
   end
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,7 @@
 source 'https://rubygems.org'
 gemspec :name => 'gollum-lib'
 gem 'irb'
+gem 'gollum-rugged_adapter', git: 'https://github.com/dometto/rugged_adapter', branch: 'improve_tree_blobs'
 
 if RUBY_PLATFORM == 'java' then
   group :development do
--- a/Gemfile
+++ b/Gemfile
@@ -1,7 +1,7 @@
 source 'https://rubygems.org'
 gemspec :name => 'gollum-lib'
 gem 'irb'
-gem 'gollum-rugged_adapter', git: 'https://github.com/dometto/rugged_adapter', branch: 'improve_tree_blobs'
+gem 'gollum-rugged_adapter', git: 'https://github.com/gollum/rugged_adapter'
 
 if RUBY_PLATFORM == 'java' then
   group :development do
--- a/Gemfile
+++ b/Gemfile
@@ -1,10 +1,12 @@
 source 'https://rubygems.org'
 gemspec :name => 'gollum-lib'
 gem 'irb'
-gem 'gollum-rugged_adapter', git: 'https://github.com/gollum/rugged_adapter'
 
 if RUBY_PLATFORM == 'java' then
+  gem 'gollum-rugged_adapter', git: 'gollum-lib_rjgit_adapter', branch: 'tree_find_blob'
   group :development do
     gem 'activesupport', '~> 6.0'
   end
+else
+  gem 'gollum-rugged_adapter', git: 'https://github.com/gollum/rugged_adapter'
 end
--- a/lib/gollum-lib/file.rb
+++ b/lib/gollum-lib/file.rb
@@ -15,14 +15,14 @@ module Gollum
     # path           - One or more String path elements to join together. `nil` values are ignored.
     # page_file_dir  - kwarg String, default: nil
     def canonical_path(*path, page_file_dir: nil)
-      prefix = Pathname.new(::File.join('/', page_file_dir.to_s))
-      rest = Pathname.new(::File.join('/', path.compact)).cleanpath
-      result = Pathname.new(::File.join(prefix, rest)).cleanpath.to_s[1..-1]
+      prefix = Pathname.new('/') + page_file_dir.to_s
+      rest = Pathname.new('/').join(*path.compact).cleanpath.to_s[1..-1]
+      result = (prefix + rest).cleanpath.to_s[1..-1]
       result.sub!(/^\/+/, '') if Gem.win_platform? # On Windows, Pathname#cleanpath will leave double slashes at the start of a path, so replace all (not just the first) leading slashes
       result
     end
       
-      # For use with self.path_match: returns true if 'query' and 'match_path' match, strictly or taking account of the following parameters:
+      # For use with self.find: returns true if 'query' and 'match_path' match, strictly or taking account of the following parameters:
       # hyphened_tags  - If true, replace spaces in match_path with hyphens.
       # case_insensitive - If true, compare query and match_path case-insensitively
       def path_compare(query, match_path, hyphened_tags, case_insensitive)
@@ -51,9 +51,8 @@ module Gollum
     # for which on_disk? is actually true.
     def self.find(wiki, path, version, try_on_disk = false, global_match = false)
       query_path =  self.canonical_path(path, page_file_dir: wiki.page_file_dir)
-      path_segments = query_path.split('/')
-      filename = path_segments.last
-      dir = path_segments[0..-2].join('/')
+      dir, filename = Pathname.new(query_path).split
+      dir = dir.to_s
 
       if global_match && self.respond_to?(:global_find) # Only implemented for Gollum::Page
         return self.global_find(wiki, version, query_path, try_on_disk) 
@@ -61,10 +60,10 @@ module Gollum
         begin
           root = wiki.commit_for(version.to_s)
           return nil unless root
-          tree = dir.empty? ? root.tree : root.tree / dir
+          tree = dir == '.' ? root.tree : root.tree / dir
           return nil unless tree
           entry = tree.find_blob do |blob_name|
-            path_compare(filename, blob_name, wiki.hyphened_tag_lookup, wiki.case_insensitive_tag_lookup)
+            path_compare(filename.to_s, blob_name, wiki.hyphened_tag_lookup, wiki.case_insensitive_tag_lookup)
           end
           entry ? self.new(wiki, entry, dir, version, try_on_disk) : nil
         rescue Gollum::Git::NoSuchShaFound
--- a/lib/gollum-lib/file.rb
+++ b/lib/gollum-lib/file.rb
@@ -7,6 +7,16 @@ module Gollum
     FS_SUPPORT_SYMLINKS = !Gem.win_platform?
 
     class << self
+
+    # Get a canonical path to a file, removing leading slashes.
+    #
+    # path           - One or more String path elements to join together
+    def canonical_path(*path)
+      # First create a clean path (e.g. no '..'), then remove leading slash.
+      result = Pathname.new(::File.join(['/', path.compact])).cleanpath.to_s
+      result.sub!(/^\/+/, '') # On Windows, Pathname#cleanpath will leave double slashes at the start of a path, so replace all (not just the first) leading slashes
+      result
+    end
       
       # For use with self.path_match: returns true if 'query' and 'match_path' match, strictly or taking account of the following parameters:
       # hyphened_tags  - If true, replace spaces in match_path with hyphens.
@@ -36,17 +46,16 @@ module Gollum
     # that if you specify try_on_disk=true, you may or may not get a file
     # for which on_disk? is actually true.
     def self.find(wiki, path, version, try_on_disk = false, global_match = false)
-      query_path = Pathname.new(::File.join(['/', wiki.page_file_dir, path].compact)).cleanpath.to_s
-      query_path.sub!(/^\/+/, '') # On Windows, Pathname#cleanpath will leave double slashes at the start of a path, so replace all (not just the first) leading slashes
-      segments = query_path.split('/')
-      filename = segments.last
-      dir = segments[0..-2].join('/')
+      path_segments =  self.canonical_path(wiki.page_file_dir, path).split('/')
+      filename = path_segments.last
+      dir = path_segments[0..-2].join('/')
+
 
       if global_match
         return nil
       else
         begin
-          root = wiki.repo.git.commit_from_ref(version.to_s)
+          root = wiki.commit_for(version.to_s)
           return nil unless root
           tree = dir.empty? ? root.tree : root.tree / dir
           return nil unless tree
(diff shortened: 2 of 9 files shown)


lib/gollum-lib/file.rb
```diff
@@ -47,18 +36,27 @@ def path_compare(query, match_path, hyphened_tags, case_insensitive)
...
-          path_match(query_path, entry, global_match, wiki.hyphened_tag_lookup, wiki.case_insensitive_tag_lookup)
+      if global_match
+        return nil
+      else
+        begin
+          root = wiki.repo.git.commit_from_ref(version.to_s)
+          return nil unless root
+          tree = dir.empty? ? root.tree : root.tree / dir
+          return nil unless tree
+          entry = tree.blobs.find do |blob|
```
dometto, line 53: This could also be faster: `tree.blobs` first wraps *all* blobs in a directory in `Gollum::Git::Blob` objects, but we only need to do this to the one that matches. So we could alter the adapters so implementing a `tree.find_blob`, which takes a block.

lib/gollum-lib/file.rb
```diff
@@ -47,18 +36,27 @@ def path_compare(query, match_path, hyphened_tags, case_insensitive)
...
+      query_path.sub!(/^\/+/, '') # On Windows, Pathname#cleanpath will leave double slashes at the start of a path, so replace all (not just the first) leading slashes
+      segments = query_path.split('/')
+      filename = segments.last
+      dir = segments[0..-2].join('/')
 
-      begin
-        entry = map.detect do |entry|
-          path_match(query_path, entry, global_match, wiki.hyphened_tag_lookup, wiki.case_insensitive_tag_lookup)
+      if global_match
+        return nil
```
dometto, line 46: This still needs to be implemented. My proposal would be to just use the old method (with the treemap) for this. Since global tag lookup  is a legacy support feature (although  I assume some people use it because they like it), it's ok if that has the slow performance.

lib/gollum-lib/page.rb
```diff
@@ -10,18 +10,23 @@ class << self
       #
       # query     - The String path to match.
       # entry     - The BlobEntry to check against.
-      # global_match - If true, find a File matching path's filename, but not its directory (so anywhere in the repo)
-      def path_match(query, entry, global_match = false, hyphened_tags = false, case_insensitive = false)
+      def global_path_match(query, entry, hyphened_tags = false, case_insensitive = false)
```
dometto, line 13: This method can be used to implement global lookup (see https://github.com/gollum/gollum-lib/pull/437#r1069787205).

lib/gollum-lib/file.rb
```diff
@@ -8,16 +8,15 @@ class File
...
-      # global_match - (Not implemented for File, see Page.path_match)
-      # hyphened_tags  - If true, replace spaces in match_path with hyphens.
-      # case_insensitive - If true, compare query and match_path case-insensitively
-      def path_match(query, entry, global_match = false, hyphened_tags = false, case_insensitive = false)
-        path_compare(query, ::File.join('/', entry.path), hyphened_tags, case_insensitive)
-      end
+    # Get a canonical path to a file, removing leading slashes.
+    #
+    # path           - One or more String path elements to join together
+    def canonical_path(*path)
```
dometto, line 14: Alternative name recommendations welcome

lib/gollum-lib/page.rb
```diff
@@ -10,18 +10,35 @@ class << self
...
-          global_match ? nil : entry.dir,
-        entry_name
+          '/', 
+          entry.dir,
+        entry.name
         ].compact)
         path_compare(query, match_path, hyphened_tags, case_insensitive)
       end
+
+      def global_find(wiki, version, query, try_on_disk)
```
dometto, line 24: This is essentially the old `Gollum::File.find` method.

dometto:
Symlinks are failing because we now use [this adapter method](https://github.com/gollum/rjgit_adapter/blob/master/lib/rjgit_adapter/git_layer_rjgit.rb#L484) to find blobs, and that method does not set the `mode` for the newly generated `Git::Blob` object (compare [here](https://github.com/gollum/rjgit_adapter/blob/master/lib/rjgit_adapter/git_layer_rjgit.rb#L217) for what needs to happen).

Gemfile
```diff
@@ -1,6 +1,7 @@
 source 'https://rubygems.org'
 gemspec :name => 'gollum-lib'
 gem 'irb'
+gem 'gollum-rugged_adapter', git: 'https://github.com/dometto/rugged_adapter', branch: 'improve_tree_blobs'
```
dometto, line 4: Needs to be removed

lib/gollum-lib/file.rb
```diff
@@ -8,16 +8,19 @@ class File
...
-      def path_match(query, entry, global_match = false, hyphened_tags = false, case_insensitive = false)
-        path_compare(query, ::File.join('/', entry.path), hyphened_tags, case_insensitive)
-      end
+    # Get a canonical path to a file.
+    # Ensures that the result is always under page_file_dir (prevents path traversal), if set.
+    # Removes leading slashes.
+    #
+    # path           - One or more String path elements to join together. `nil` values are ignored.
+    # page_file_dir  - kwarg String, default: nil
+    def canonical_path(*path, page_file_dir: nil)
```
dometto, line 17: I am wondering whether it makes more sense for this code to live in the `GitAccess` class.

lib/gollum-lib/file.rb
```diff
@@ -41,24 +44,32 @@ def path_compare(query, match_path, hyphened_tags, case_insensitive)
...
+      filename = path_segments.last
+      dir = path_segments[0..-2].join('/')
 
-      begin
-        entry = map.detect do |entry|
-          path_match(query_path, entry, global_match, wiki.hyphened_tag_lookup, wiki.case_insensitive_tag_lookup)
+      if global_match && self.respond_to?(:global_find) # Only implemented for Gollum::Page
+        return self.global_find(wiki, version, query_path, try_on_disk) 
+      else
+        begin
```
dometto, line 61: Similarly, I am thinking it may make sense to refactor this code so that it mostly becomes a method of `GitAccess` (so that we would hear call e.g. `wiki.access.find_blob_in_tree(dir, filename) {|blob_name| path_compare(filename, blob_name, ...}`.

dometto:
I've written a small script that benchmarks the performance of reading (getting `formatted_data`) and writing pages on a) a wiki with many commits and b) a wiki with many blobs and trees. See [gist](https://gist.github.com/dometto/2a8ec37b48bf814b115b4b16861a7ed7).

## Results:

### On `master`

===== Repo with many trees and blobs
Reading pages
                 user     system      total        real
            20.933988   0.359543  21.293531 ( 22.686614)
Writing pages
                 user     system      total        real
             1.825870   0.224374   2.050244 (  2.227173)
===== Repo with many commits
Reading pages
                 user     system      total        real
             0.347461   0.014023   0.361484 (  0.380127)
Writing pages
                 user     system      total        real
             0.318842   0.433577   0.752419 (  1.229406)

### On this branch

===== Repo with many trees and blobs
Reading pages
                 user     system      total        real
             0.124437   0.027632   0.152069 (  0.183316)
Writing pages
                 user     system      total        real
             1.842885   0.350508   2.193393 (  2.913138)
===== Repo with many commits
Reading pages
                 user     system      total        real
             0.303247   0.027656   0.330903 (  0.344811)
Writing pages
                 user     system      total        real
             0.266384   0.352682   0.619066 (  0.777089)

dometto:
As we can see the changes on this branch improve performance drastically, except for the case of writing files, for some reason. ☹️ I didn't think I touched any code related to the latter, but: since the time increase for writing pages is only small (keep in mind the benchmark runs the write 5 for two pages 5 times), I think the benefits clearly outweigh the damage here. :)

lib/gollum-lib/file.rb
dometto: @bartkamphorst what do you think about this?

lib/gollum-lib/file.rb
```diff
@@ -41,24 +44,32 @@ def path_compare(query, match_path, hyphened_tags, case_insensitive)
...
+          root = wiki.commit_for(version.to_s)
```
bartkamphorst, line 62: What kind of data structure is `version` that it needs the `to_s` call?

lib/gollum-lib/file.rb
```diff
@@ -41,24 +44,32 @@ def path_compare(query, match_path, hyphened_tags, case_insensitive)
...
+          return nil unless root
+          tree = dir.empty? ? root.tree : root.tree / dir
+          return nil unless tree
```
bartkamphorst, line 65: Just as a sanity check: can `root.tree` ever be `nil`? Or can `root.tree / dir` be `nil`? Otherwise this check might not be needed.

lib/gollum-lib/git_access.rb
```diff
@@ -161,15 +161,11 @@ def tree!(sha)
       items = []
       tree.each do |entry|
         if entry[:type] == 'blob'
-          items << BlobEntry.new(entry[:sha], entry[:path], entry[:size], entry[:mode].to_i(8))
+          next if @page_file_dir && !entry[:path].start_with?("#{@page_file_dir}/")
```
bartkamphorst, line 164: Seems like a nicer solution. 👍

lib/gollum-lib/page.rb
```diff
@@ -10,18 +10,35 @@ class << self
...
         return false if "#{entry.name}".empty?
         return false unless valid_extension?(entry.name)
-        entry_name = valid_extension?(query) ? entry.name : strip_filename(entry.name)     
         match_path = ::File.join([
-          '/',
-          global_match ? nil : entry.dir,
-        entry_name
+          '/', 
+          entry.dir,
+        entry.name
```
bartkamphorst, line 19: I'm sure there's a Pathname variant you could use here, but I'm not sure it would be more elegant. In terms of consistency, you might want to consider it (e.g., `Pathname.new('/').join([etc. ...])`).

bartkamphorst (changes requested):

lib/gollum-lib/file.rb
dometto: It's a `Gollum::Git::Commit`. We could use `version.id` (which is an alias of `to_s`) instead?

dometto: Actually, according to the inline docs it should be a `String`. We may be assuming that it *can* also be a Commit, in which case `to_s` is an option that works in either case. What do you think is better?

lib/gollum-lib/file.rb
dometto: If `version` can only be a Commit (see other comment), then presumably it will never be `nil`. If Commit is/can be a String, `version` could be `nil`.

lib/gollum-lib/file.rb
```diff
@@ -8,16 +8,19 @@ class File
...
-      # query     - The String path to match.
-      # entry     - The BlobEntry to check against.
-      # global_match - (Not implemented for File, see Page.path_match)
-      # hyphened_tags  - If true, replace spaces in match_path with hyphens.
-      # case_insensitive - If true, compare query and match_path case-insensitively
-      def path_match(query, entry, global_match = false, hyphened_tags = false, case_insensitive = false)
-        path_compare(query, ::File.join('/', entry.path), hyphened_tags, case_insensitive)
-      end
+    # Get a canonical path to a file.
+    # Ensures that the result is always under page_file_dir (prevents path traversal), if set.
```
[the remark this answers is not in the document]
dometto: If I remember my own code correctly, yes, but please sanity-check me! It should be being tested here:

 https://github.com/gollum/gollum-lib/pull/437/files#diff-a791cacf8158fbb59f8b0acb0545ebdbbd989ca95115a99fcab5585bd1ba1276R14

lib/gollum-lib/file.rb
dometto: `canonical_path` is currently a Class method, and needs to be that because we are using it when looking up files (i.e. when there is no instance of a `File` yet). We also use the same method once in `File#initialize`, but there the result is already cached. So the question is whether `canonical_path` makes more sense as a Class method on `File`, or on `GitAccess`?

lib/gollum-lib/file.rb
```diff
@@ -41,24 +44,32 @@ def path_compare(query, match_path, hyphened_tags, case_insensitive)
...
     #
     # Returns a Gollum::File or nil if the file could not be found. Note
     # that if you specify try_on_disk=true, you may or may not get a file
     # for which on_disk? is actually true.
     def self.find(wiki, path, version, try_on_disk = false, global_match = false)
-      map = wiki.tree_map_for(version.to_s)
-
-      query_path = Pathname.new(::File.join(['/', wiki.page_file_dir, path].compact)).cleanpath.to_s
-      query_path.sub!(/^\/\//, '/') if Gem.win_platform? # On Windows, Pathname#cleanpath will leave double slashes at the start of a path intact, so sub them out.
+      query_path =  self.canonical_path(path, page_file_dir: wiki.page_file_dir)
```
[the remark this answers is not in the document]
dometto: We're in a class method here, so `self.canonical_path` is not referring to an instance of a `File` -- so we can't cache for any particular instance. The logic here is: we're looping over the repository trying to find a blob whose path in the repository matches `page_file_dir + path`. But since we can't just add those together because of possible path traversal exploits (if the user sets `path` as `../../`), we have to use the `canonical_path` to join them together.

lib/gollum-lib/file.rb
dometto: I've opened [this issue](https://github.com/gollum/gollum/issues/1972) to resolve the ambiguity.

lib/gollum-lib/file.rb
dometto: It looks like `root.tree / dir` can be `nil`, when the user looks for a non-existent path. So best to keep the `nil` check.

lib/gollum-lib/file.rb
```diff
@@ -41,24 +44,32 @@ def path_compare(query, match_path, hyphened_tags, case_insensitive)
...
+      path_segments = query_path.split('/')
```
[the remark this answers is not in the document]
dometto: Great, thanks!

lib/gollum-lib/file.rb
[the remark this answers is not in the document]
dometto: Would a better name for `File.canonical_path` be `File.cleanpath`? It's perhaps more descriptive, but invites confusion with `Pathname.cleanpath.`

lib/gollum-lib/page.rb
dometto: The working `Pathname` variant would be:

```ruby
        match_path = Pathname.new('/').join(*[
          entry.dir,
          entry.name
          ].compact
        ).to_s
```
But since it adds a splat and a `.to_s` call, I'm also not sure if it's more elegant. I'll leave it as is for now, if you think consistency is important here I'm happy to go for that.

bartkamphorst: My mission is consistency in the use of Pathname, but this is your call.

lib/gollum-lib/file.rb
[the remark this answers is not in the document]
bartkamphorst: I'm on board with `canonical_path`.

lib/gollum-lib/page.rb
dometto: Went for the `Pathname` variant :)

lib/gollum-lib/file.rb
dometto: This turns out to be more complicated than I thought, as there is currently no `attr_accessor` for `wiki.access`. So either we have to make `wiki.access` public, or the `find_blob` method should be implemented as an instance method on `Wiki` that itself calls the underlying `GitAccess` instance, as e.g. `Wiki#tree_list` already does. Any opinion about what would make for the best API?

dometto: As additional info, the call in `file.rb` would look something like this:

```
        entry = wiki.access.find_blob_in_tree(wiki.commit_for(version.to_s), dir.to_s) do |blob_name|
```

Where `wiki.commit_for` itself invokes the `GitAccess` class. This feels a bit ugly.

dometto: After discussion we've concluded this change is too complicated and not worth it, so we're leaving the find file logic in`file.rb` as opposed to `git_access.rb`

PR #439, opened 2023-01-18
Fix Rakefile release task

dometto:
The release flow failed for me because `git tag` suddenly asked for a message. Is it possible the default for `git tag` is now to require a tag message (rather than using "lightweight tags" by default)? If anyone knows anything about this I'd be much obliged. :)

In this PR I've fixed it by passing a message along with `git tag -m`.

Note: this should also be fixed in gollum.

### DIFF
--- a/LATEST_CHANGES.md
+++ b/LATEST_CHANGES.md
@@ -1,4 +1,4 @@
 # 5.2.2 / 2023-01-18
 
-* Bugfix release: set Nokogiri default XHTML conversion options more relaibly. See https://github.com/sparklemotion/nokogiri/issues/2761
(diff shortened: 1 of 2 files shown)


dometto:
Also the automatic github release workflow failed because `LATEST_CHANGES.md` wasn't committed, so I include it here. I've manually updated the GitHub release.

LATEST_CHANGES.md
```diff
@@ -0,0 +1,4 @@
+# 5.2.2 / 2023-01-18
+
+* Bugfix release: set Nokogiri default XHTML conversion options more relaibly. See https://github.com/sparklemotion/nokogiri/issues/2761
```
bartkamphorst, line 3: Typo: reliably.

PR #440, opened 2023-01-27
Abstract use of Octicons to Gollum::Icon class

bartkamphorst:
At present, the `octicons` gem is used directly in both `gollum-lib` and `gollum`. This PR is a first step to streamline this use across both gems by way of a `Gollum::Icon` class with a `get_icon` method. Another PR in gollum will follow at a later stage. This PR does two further things:

* It relaxes the constraints on the `octicons` gem version in gollum-lib. This ensures that `gollum-lib` will still have some version of that gem available when `gollum-lib` is used without `gollum`, without getting in the way of upgrading `octicons` at the frontend side  (`gollum`) where the version constraints are more strict.
* It ensures that if a specific octicon is not available, a fallback icon is used.

### DIFF
--- a/lib/gollum-lib/icon.rb
+++ b/lib/gollum-lib/icon.rb
@@ -1,0 +1,33 @@
+module Gollum
+  class Icon
+    # See if octicons gem is available
+    @@octicons = false
+    begin
+      require 'octicons'
+      @@octicons = true
+    rescue LoadError; end
+    
+    def self.get_icon(name, options={}, for_css=false)
+      defaults = {height: 16, width: 16, class: ''}
+      options = defaults.merge(options)
+      if for_css
+        options.delete(:height)
+        options.delete(:width)
+      end
+      if @@octicons && octicon = ::Octicons::Octicon.new(name, options)
+        octicon.to_svg
+      else
+        height = %Q(height="#{options[:height]}") if options[:height]
+        width = %Q(width="#{options[:width]}") if options[:width]
+        cls = %Q(class="octicon octicon-alert #{options[:class]}")
+        xmlns = %Q(xmlns="http://www.w3.org/2000/svg") if for_css
+        title = %Q(<title>Octicons are at present not available</title>)
+        viewbox = %Q(viewBox="0 0 24 24") if options[:width]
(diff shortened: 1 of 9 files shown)


bartkamphorst:
@benjaminwil Curious to hear your take on this approach as well.

bartkamphorst:
JRuby failure seems unrelated to this PR.

lib/gollum-lib/icon.rb
```diff
@@ -0,0 +1,33 @@
...
+      defaults = {height: 16, width: 16, class: ''}
+      options = defaults.merge(options)
+      if for_css
+        options.delete(:height)
+        options.delete(:width)
+      end
+      if @@octicons && octicon = ::Octicons::Octicon.new(name, options)
+        octicon.to_svg
+      else
+        height = %Q(height="#{options[:height]}") if options[:height]
```
dometto, line 20: The implementation of this class looks good to me! I would suggest abstracting this else block into a separate method, `render_default_icon(name, options)` or some such.

bartkamphorst: Done. 👍

benjaminwil:
I like this change a lot. I do wonder if it would be possible to drop the Octicons dependency entirely, though.

My impression of `gollum-lib` (though I'm not that familiar with this codebase yet) is that it isn't for presentational code.

What if we used the CSS classes on the flash `<div>`s to inject any required icons.

So in `gollum-lib`:

```html
<div class="flash flash-error"> # etc...
```

and in Gollum, something like:

```css
.flash.flash-error::before {
  content: url('path-to-the-svg.svg');

  /* Styles below to replicate the octicons class. */
  position: relative;
  top: 4px;
  margin-right: 16px;
}
```

The tradeoff is that we would need to manage special icon classes in Gollum's CSS.

But the benefit would be including one fewer dependencies in `gollum-lib` and having it care less about document presentation in general.

I made a quick proof of concept to demonstrate the icon injection:

https://codepen.io/benjaminwil/pen/vYarjQm

bartkamphorst:
Neat idea @benjaminwil ; certainly a possibility that warrants further consideration. What I'd like to retain are the `Note`, `Warn` and `Octicon` macros, which live in `gollum-lib`. But what if we were to embed the name of the requested icon (e.g., 'bell') and options (`{width: 16, height: 16}`) as data-* attributes to the respective divs these macros output? Then, I think, it would suffice to only maintain one icon styling directive, not multiple. Correct?

benjaminwil:
That sounds like a good strategy. I would merge that.

My argument against that strategy would be that that this data is more suited to live in Gollum, where the rest of the icons and styles live.

The icon name and dimensions are only a concern of Gollum, which presents the icons. It may result in a few more lines of code, but I think it gives us flexibility long term. For example: if we decided to use a different icons library in Gollum, we would need to make changes in `gollum` and `gollum-lib` again. If Gollum is the sole provider of icon logic, we can save ourselves the trouble of updating two gems.

bartkamphorst:
> If Gollum is the sole provider of icon logic, we can save ourselves the trouble of updating two gems.

I think we are broadly in agreement. It's just that the `(Oct)Icon` macro (and the `Note` macro for that matter) is expanded inside gollum-lib. So if we want to keep these macros that allow users to insert some selected icon (`Icon`) and insert flash statements with a "custom" icon (`Note`), we need a way to feed the requested icon back to gollum. My proposal was to do that through data-* attributes. Have a look at #441 to see what I have in mind.

bartkamphorst:
Superseded by #441 .

PR #442, opened 2023-02-09
`<<Video ..., auto=true>>` to auto-play inline video (GIF style)

aljungberg:
- Using the `="true"` work-around as in #430.
- The change in `sanitization.rb` further addresses gollum/gollum#1587 by allowing more of the standard video properties.
- Also featured: less escaping, capitalisation fix.

### DIFF
--- a/lib/gollum-lib/macro/video.rb
+++ b/lib/gollum-lib/macro/video.rb
@@ -1,8 +1,10 @@
 module Gollum
   class Macro
     class Video < Gollum::Macro
-      def render (fname)
-        "<video width=\"100%\" height=\"100%\" src=\"#{CGI::escapeHTML(fname)}\" controls=\"true\"> HTML5 video is not supported on this Browser.</video>"
+      def render(fname, auto=false)
+        escaped_fname = CGI.escapeHTML(fname)
+        properties = auto ? "autoplay='true' playsinline='true' muted='true' loop='true'" : "controls='true'"
+        "<video width='100%' height='100%' src='#{escaped_fname}' #{properties}>HTML5 video is not supported on this browser.</video>"
       end
     end
   end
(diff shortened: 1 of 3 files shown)


dometto:
Thanks again! Sadly it looks like the JRuby tests fail because the markdown parser on JRuby generates `<p>` tags around the video tag. I think this is nothing to worry about, you could just make the expectation regex more lenient.

aljungberg:
Hmm I’m not sure, isn’t the regex a find rather than a match? It feels like the first test should have failed the same way if extra `<p>`s were the problem, but it seems only the second test is failing. Also looking for that space to `%20` escape and I’m not seeing that in the actual output, instead the space seems to have been silently swallowed. Something different about how JRuby parses the macro tag such that spaces in filenames get eaten?

```
</<video (.*)src="\/Uploads\/foo%20.mp4"(.*)>(.*)<\/video>/> was expected to be =~
[17](https://github.com/gollum/gollum-lib/actions/runs/4436685631/jobs/7788167501?pr=442#step:6:18)
  <"<p><video autoplay=\"true\" height=\"100%\" loop=\"true\" muted=\"true\" playsinline=\"true\" src=\"/Uploads/foo .mp4\" width=\"100%\">HTML5 video is not supported on this browser.</video></p>\n">.
```

dometto:
Ah, you're right about the location of the problem! It seems that `CGI.escapeHTML` does not convert spaces to `%20` on either JRuby or MRI. So the `%20` on MRI must be getting produced somewhere else in gollum, probably due to a difference in the rendering gems between MRI./JRuby. Maybe we can just change the example filename to test escaping, e.g. to something like `/Uploads/file<script>.mp4`? 

```ruby
jruby-9.4.0.0 :004 > CGI.escapeHTML("/Uploads/file<script>.mp4")
 => "/Uploads/file&lt;script&gt;.mp4" 
```

aljungberg:
Yep works for me! Pushed.

dometto:
Thanks for your effort!

PR #447, opened 2023-03-26
Ensure boolean Macro arguments are parsed as boolean. Resolves #441

dometto:
See #441. I have rewritten `Macro#extract` to handle `false` and `true` arguments explicitly.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -6,7 +6,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        ruby: [jruby-9.3.2.0]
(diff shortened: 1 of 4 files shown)


bartkamphorst (approved):
Looks clean. Okay to merge as far as I'm concerned.

dometto:
FYI, I've had to drop support for Ruby 2.6 and JRuby 9.3 (which is Ruby 2.6) to get the tests passing, as this PR utilizes the `case .. in` syntax (pattern matching) introduced in Ruby 2.7. JRuby 9.4 has been out since 2022-11-23 -- not terribly long, do we want to keep support for 9.3?

PR #452, opened 2023-08-01
Fix absolute path includes from a subdir

dometto:
Resolves https://github.com/gollum/gollum/issues/1978

As https://github.com/gollum/gollum/issues/1978 explains, include tags with an absolute path currently do not work. The file-to-include is always resolved as the current directory *plus* the path provided in the tag. This does not work when attempting to include an absolute path in a document in a subidr (the resolved file will then be `"/subdir/absolute/path"` instead of `"/absolute/path"`).

This PR addresses the issue by checking the include path in order to determine whether it is absolute or not, and only prepending the current directory path in case it is relative.

Many thanks to @aaakbar for identifying the issue and pinpointing the solution!

### DIFF
--- a/lib/gollum-lib/filter/tags.rb
+++ b/lib/gollum-lib/filter/tags.rb
@@ -120,8 +120,12 @@ class Gollum::Filter::Tags < Gollum::Filter
   def process_include_tag(tag)
     len = INCLUDE_TAG.length
     return html_error('Cannot process include directive: no page name given') if tag.length <= len
-    page_name          = tag[len..-1]
-    resolved_page_name = ::File.join(@markup.dir, page_name)
+    page_path = Pathname.new(tag[len..-1])
+    if page_path.relative?
+      resolved_page_name = (Pathname.new(@markup.dir) + page_path).to_s
+    else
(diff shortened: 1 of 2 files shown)


lib/gollum-lib/filter/tags.rb
```diff
@@ -120,8 +120,12 @@ def parse_tag_parts(tag)
   def process_include_tag(tag)
     len = INCLUDE_TAG.length
     return html_error('Cannot process include directive: no page name given') if tag.length <= len
-    page_name          = tag[len..-1]
-    resolved_page_name = ::File.join(@markup.dir, page_name)
+    page_path = Pathname.new(tag[len..-1])
```
bartkamphorst, line 123: How about:
```ruby
page_path = Pathname.new(tag[len..-1])
resolved_page = page_path.relative? ? (Pathname.new(@markup.dir) + page_path) : page_path
resolved_page_name = resolved_page.cleanpath.to_s

```

dometto: Thanks, much nicer!

bartkamphorst (approved):
👍

PR #454, opened 2023-12-27
HTML escape YAML after parsing to prevent invalidating YAML string

dometto:
Resolves https://github.com/gollum/gollum/issues/2024

We were previously HTML escaping the entire YAML string before parsing it. This means that some valid YAML was being scrubbed (in the case of the reported bug, the `>` character, but who knows if there are more?). This new approach first parses the YAML, and then recursively HTML escapes every key and value in the resulting hash.

### DIFF
--- a/lib/gollum-lib/filter/yaml.rb
+++ b/lib/gollum-lib/filter/yaml.rb
@@ -10,7 +10,7 @@ class Gollum::Filter::YAML < Gollum::Filter
     data.gsub!(YAML_FRONT_MATTER_REGEXP) do
       @markup.metadata ||= {}
       begin
-        frontmatter = ::YAML.safe_load(sanitize(Regexp.last_match[1]))
+        frontmatter = sanitize_frontmatter(::YAML.safe_load(Regexp.last_match[1]))
         @markup.metadata.merge!(frontmatter) if frontmatter.respond_to?(:keys) && frontmatter.respond_to?(:values)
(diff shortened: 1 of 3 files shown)


dometto:
A (obviously faster) alternative is to *not* HTML escape/sanitize the frontmatter hash at all, and make the frontend  (mustache) responsible for escaping. @bartkamphorst what makes more sense do you think?

benjaminwil (approved):

bartkamphorst:
Implementation looks clean. Can we expect a performance hit for recursively sanetizing medium-sized YAML frontmatter?

dometto:
In theory yes, but in practice probably not? I think it is unlikely that front matter would consist of very deeply nested dicts and arrays. As a design question I'm unsure what the proper place for escaping is though (here or in the frontend). We sanitize most macro and page data in `gollum-lib`, but that's because in most cases we want to allow certain html tags and disallow others (for which we use `loofah`). For the front matter, we could (and maybe should?) just escape all html. So we could

a) let this macro return unescaped content and let mustache handle it in the frontend (a break with other macros).
Or
b) handle escaping as in this PR, but instead of using the standard Sanitization class (i.e. `loofah`), use something presumably faster like `CGI.escape_html`?

dometto:
@bartkamphorst just noticed this PR again, it's been a while but do you have any thoughts on the proper way to handle this?

bartkamphorst (approved):
I think this solution is fine!

PR #455, opened 2024-05-05
Remove `Gollum::FileView`

benjaminwil:
It appears this file is never used anywhere by `gollum` or `gollum-lib`. So we can safely remove it.

I manually verified that removing this breaks nothing:

- I manually visited all Gollum pages.
- I manually deleted a page and ensured that there was still a confirmation dialog.
- I ran the `gollum` test suite.
- I ran the `gollum-lib` test suite.

If you can think of anything else we should do to verify this change let me know. But it really seems like this is just dead code.


dometto:
Good find!

dometto (approved):

PR #456, opened 2024-07-19
Add support for GitLab style plantuml blocks and additional diagram types

th-yoo:
- Support for [GitLab style ```plantuml blocks](https://github.com/gollum/gollum/issues/1260)
- Extends the diagram types available to include [{ebnf, regex, ditaa, gantt, chronology, chen}](https://plantuml.com/).  

### Examples
- PlantUML tags are used in the same way as before
```
    @startuml
    A <|-- B
    @enduml
```

-  Defaults to `@startuml`
```
    ```plantuml
    A <|-- B
    ```
```

```
    ```plantuml

### DIFF
--- a/lib/gollum-lib/filter/code.rb
+++ b/lib/gollum-lib/filter/code.rb
@@ -47,8 +47,23 @@ class Gollum::Filter::Code < Gollum::Filter
     end
     
 
+    def replace_plantuml(data)
+      unless data.match(/\s*@start(uml|json|yaml|ebnf|regex|salt|ditaa|gantt|chronology|mindmap|wbs|math|latex|chen)/m)
+        data = "@startuml\n#{data}\n@enduml\n"
+      end
+      data
+    end
+
+    # print the SHA1 ID with the proper indentation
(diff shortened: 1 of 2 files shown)


guillaume-d:
Couldn't this be implemented using Gollum 6's [custom language handlers](https://github.com/gollum/gollum/wiki/Custom-language-handlers) instead, or am I missing something?

Also the `@start`... syntax could stay without support for new diagram types. Not adding them there may be both good (backward-compatibility) and bad (inconsistent with backtick syntax).

dometto:
Thank you for submitting this @th-yoo!

> Couldn't this be implemented using Gollum 6's [custom language handlers](https://github.com/gollum/gollum/wiki/Custom-language-handlers) instead, or am I missing something?

Thanks for mentioning this @guillaume-d -- I think that would indeed be the best and most maintainable way to implement this functionality.

@th-yoo do you think you could rework this PR to incorporate this idea? Let me know if you need pointers to implement the language handler. It might (but also might not) make sense to implement this in gollum instead of gollum-lib, compare the [Mermaid implementation](https://github.com/gollum/gollum/blob/master/lib/gollum/app.rb#L40) here.

jgsuess:
Any chance this will eventually merge? Seems like the best is the enemy of the good here... :(

PR #457, opened 2025-01-11
Improve performance by supporting repo.lstree with a specific path

dometto:
This PR adds support for new adapter features which allow us to call `lstree` for a specific directory in the repo, and non-recursively:

https://github.com/gollum/rugged_adapter/pull/67
https://github.com/gollum/rjgit_adapter/pull/26
Specs: https://github.com/gollum/adapter_specs/pull/19

This will eventually allow us to improve performance of the Overview route, as we'll no longer need to get and loop over the entire contents of the repository.

### DIFF
--- a/lib/gollum-lib/git_access.rb
+++ b/lib/gollum-lib/git_access.rb
@@ -48,9 +48,9 @@ module Gollum
     # ref - A String Git reference or Git SHA to a commit.
     #
     # Returns an Array of BlobEntry instances.
-    def tree(ref)
+    def tree(ref, path = '/', recursive = true)
       if (sha = ref_to_sha(ref))
-        get_cache(:tree, sha) { tree!(sha) }
+        get_cache(:tree, "#{sha}#{path}") { tree!(sha, path, recursive) }
       else
         []
       end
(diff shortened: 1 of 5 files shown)


lib/gollum-lib/git_access.rb
```diff
@@ -48,9 +48,9 @@ def ref_to_sha(ref)
     # ref - A String Git reference or Git SHA to a commit.
     #
     # Returns an Array of BlobEntry instances.
-    def tree(ref)
+    def tree(ref, path = '/', recursive = true)
       if (sha = ref_to_sha(ref))
-        get_cache(:tree, sha) { tree!(sha) }
+        get_cache(:tree, "#{sha}#{path}") { tree!(sha, path, recursive) }
```
dometto, line 53: Cache key currently doesn't differentiate between `recursive` being true or false.

lib/gollum-lib/git_access.rb
```diff
@@ -156,16 +156,19 @@ def ref_to_sha!(ref)
...
-      tree  = @repo.lstree(sha, { :recursive => true })
-      items = []
-      tree.each do |entry|
-        if entry[:type] == 'blob'
-          next if @page_file_dir && !entry[:path].start_with?("#{@page_file_dir}/")
-          items << BlobEntry.new(entry[:sha], entry[:path], entry[:size], entry[:mode])
+    def tree!(sha, path = '/', recursive = true)
+      tree  = @repo.lstree(sha, (Pathname.new(@page_file_dir.to_s) + path).to_s, recursive: recursive )
+      tree.reduce([]) do |accumulator, entry|
+        if !@page_file_dir || entry[:path].start_with?("#{@page_file_dir}/") # guard against directory traversal
```
dometto, line 162: Not sure this line is needed, if we're already passing `@page_file_dir` as the base path.

lib/gollum-lib/git_access.rb
```diff
@@ -156,16 +156,19 @@ def ref_to_sha!(ref)
...
+          accumulator << case entry[:type]
+          when 'blob'
+            BlobEntry.new(entry[:sha], entry[:path], entry[:size], entry[:mode])
+          when 'tree'
+            TreeEntry.new(entry[:sha], entry[:path])
+          end
         end
```
dometto, line 169: We could also opt to return blobs and trees in separate Arrays, so callers who need only the blobs can ignore the trees.

Either way, I guess initializing the TreeEntries has minor performance impact when looping into the entire wiki.

PR #458, opened 2025-05-30
Anon user email at a RFC2606 reserved domain

jmtd:
The previous value "anon.com" is a real-world domain, unrelated to Gollum (or deployments of Gollum). Use a domain specified as reserved by RFC2606 and so guaranteed not to be in legitimate use elsewhere.

Fixes https://github.com/gollum/gollum/issues/2109.


dometto:
Thanks @jmtd!

PR #462, opened 2025-10-26
Require ostruct since it is not part of default ruby 2.5

fliiiix:
A lot of base libs are moved to gems with Ruby 2.5.0.

```
warning: ostruct is not part of the default gems since Ruby 3.5.0. Install ostruct from RubyGems.
```


fliiiix:
I guess since it is still in `Ruby 3.5.0-preview1` preview im not sure if it makes sense to officially support it yet, but definitely worth checking with the next release 

(I'm running gollum on FreeBSD on the 3.5.0 preview right now)

PR #464, opened 2026-04-17
allow macros to provide javascript that is added to the page

morus:
addresses [#2169](/gollum/gollum/issues/#2169)

The change turned out to be more complicated than I thought. I naively assumed that the sanitization of macro output would happen on the macro output. But that's not the case. The whole page is sanitized at the end of processing.

That makes it more complicated and less local to allow macros to provided javascript without being sanitzed.

The solution I choose is not too nice but hopefully acceptable.
A nicer solution would require to change the data passed through the filter chain from a simple string to a more structured and extendable object. That would require to touch all of the existing filters, which is beyond the amount of changes I want to do.

My change allows filters to collect a set of script texts (strings) in an instance variable of the filter and builds a union of scripts when the filters are run (in markup.rb) that is added to the output within script tags.

### DIFF
--- a/lib/gollum-lib/filter.rb
+++ b/lib/gollum-lib/filter.rb
@@ -49,6 +49,8 @@ module Gollum
 
     PLACEHOLDER_PATTERN = /%(\S+)%.+=\1=/
 
+    attr_reader :scripts
+
     # Setup the object.  Sets `@markup` to be the instance of Gollum::Markup that
     # is running this filter chain, and sets `@map` to be an empty hash (for use
     # in your extract/process operations).
@@ -58,6 +60,7 @@ module Gollum
       @map    = {}
       @open_pattern = "%#{self.class.to_s.split('::').last}%"
       @close_pattern = "=#{self.class.to_s.split('::').last}="
+      @scripts = Set.new
     end
 
     attr_reader :open_pattern, :close_pattern
@@ -70,6 +73,10 @@ module Gollum
     def process(data)
       raise RuntimeError,
             "#{self.class} has not implemented ##process!"
+    end
+
+    def add_javascript(script)
+      @scripts.add(script) if script
     end
 
     protected
--- a/lib/gollum-lib/filter/macro.rb
+++ b/lib/gollum-lib/filter/macro.rb
@@ -51,7 +51,9 @@ class Gollum::Filter::Macro < Gollum::Filter
 
       data.gsub!(id) do
         begin
-          Gollum::Macro.instance(macro, @markup.wiki, @markup.page).render(*args)
+          macro_instance = Gollum::Macro.instance(macro, @markup.wiki, @markup.page)
+          add_javascript(macro_instance.javascript)
+          macro_instance.render(*args)
         rescue StandardError => e
           %Q(<div class="flash flash-error gollum-macro-error my-2">Macro Error for #{macro}: #{e.message}</div>)
         end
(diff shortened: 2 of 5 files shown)


dometto:
Thanks for the PR! I understand the difficulty and think the current strategy actually looks pretty clean. Pinging @bartkamphorst or @benjaminwil for a second opinion since this is security-related :)

morus:
There's a test issue with ruby 3.5 (3.3 runs tests successfully but fails then). Does not seem to be related to my change. Maybe the issue is, that there is no ruby 3.5 (rbenv lists a 3.5-preview1, but this seems to continue as 4.0).
I ran specs locally with ruby 3.0.5, 3.4.5 and 4.0.0 without issues (3.0.5 was the oldest version I have installed here; there were some gem issues (guard dependencies) that did not affect the tests)

Not sure what ruby versions should be supported. The gemspec says >= 2.4. But that's very old, support ended in 2020, support for 3.0 ended in 2024.

Not being used to gem development, I admittedly didn't think much about old ruby versions.
In #465 I had to change a use of the Data class into a Struct to make it work with 3.0.

dometto:
On 3.5 it's complaining about nokogiri failing to install -- that's odd. I'm happy to drop support for 3.5 if other 3.x rubies, and 4.0, work. I've opened a new PR to test against 4.0.2. Let's see what that does.

benjaminwil:
This looks like a good and safe solution. Wiki administrators must trust anyone they allow as a committer, anyway. So a wiki's trusted users gain a bit of power/flexibility, and untrusted users are still unable to be malicious when inserting macros. I like it.

This is an impactful enough change that I think it warrants a minor release. 😄 

An aside: in `gollum`, it could be good to add explicit tests so that we're certain that users cannot add arbitrary macros containing malicious `#script` contents (or any Ruby files, really) via the Gollum editor. I am pretty confident that that _is_ how Gollum works, but I am less confident that that's under test.

morus:
Thanks for the reviews.

Security wise I'd say one has to distinguish between Admins (running gollum and being able to add macros) and Users.

Admins can (before and after the change) write macros that allow arbitrary ruby code execution (e.g. a macro with a string parameters that calls eval on that string). Not reasonable but possible.
With the change, they can also write macros that allow arbitrary javascript addition in the page output.

When admins add such macros, users are able to use and abuse them. It's the responsibility of the admins to not do that (unless they trust the users to not abuse possibilities; in my case I'll use gollum for personal knowledge management and could hopefully trust me as a user).

Macros added to gollum-lib should probably be not allowed to have dynamic stuff in their javascript, but that's rather a review rule. And the static JS should not do dangerous stuff...

The section in the gollum wiki regarding custom macros should probably contain a warning about abuse possibilities of dynamic javascript that includes wiki user content (and/or there could be a comment on the empty javascript method in the macro base class).
I probably would not talk about the ruby code execution as it's to far fetched anyways.

Regarding specs that check that that gollum does not execute wiki content, I'm not sure how one would write that.
Ok, add a ruby script as wiki content, but then you'd have to run all potential ways how that code might be executed.
Regarding JS there is a test in test_markup.rb that script tags are removed.

dometto:
A rebase on `master` should fix the test issues

benjaminwil:
> Regarding specs that check that that gollum does not execute wiki content, I'm not sure how one would write that.
> Ok, add a ruby script as wiki content, but then you'd have to run all potential ways how that code might be executed.

My thought was that we simply do not want users to be able to add a `.rb` file _at all_ from the Gollum editor. Or said another way: only users with access to the repository should be able to add files containing usable macros.

dometto:
I've created https://github.com/gollum/gollum/issues/2171 to track @benjaminwil's suggestion (with a specific implementation in mind -- feel free to change).

PR #465, opened 2026-04-19
structured global toc macro

morus:
Adds a "structured global toc" macro that creates a list of all pages with their folder structure.

Note that the macro does not fully work without #464 (alternatively one can provide the javascript code as an external javascript). Opening and closing folders will not be functional.

### DIFF
--- a/gemspec.rb
+++ b/gemspec.rb
@@ -95,6 +95,7 @@ def specification(version, default_adapter, platform = nil)
     lib/gollum-lib/macro/navigation.rb
     lib/gollum-lib/macro/note.rb
     lib/gollum-lib/macro/series.rb
+    lib/gollum-lib/macro/structured_global_toc.rb
     lib/gollum-lib/macro/video.rb
     lib/gollum-lib/macro/warn.rb
     lib/gollum-lib/markup.rb
--- a/lib/gollum-lib/macro/structured_global_toc.rb
+++ b/lib/gollum-lib/macro/structured_global_toc.rb
@@ -1,0 +1,132 @@
+# frozen_string_literal: true
+
+module Gollum
+  class Macro
+    class StructuredGlobalTOC < Gollum::Macro
+      TocPage = Struct.new(:page, :path) do
+        def initialize(page:, path: nil)
+          super(page, path || page.url_path.split('/'))
+        end
+      end
+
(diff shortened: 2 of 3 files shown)


dometto:
This looks cool! However I'm a bit unsure on whether to add JavaScript to the gollum-lib repo, for maintenance reasons. This is for sure a matter of degree, as we're already keeping some html in the filters in this repo. But it would be good to limit this as much as possible. I guess it comes down to: how badly do we want this Macro to be part of gollum standardly?

An alternative could be to package it in `contrib` in the Gollum repo.

@benjaminwil @bartkamphorst happy to hear what you think.

bartkamphorst:
I agree with wanting to limit the amount of JS in `gollum-lib` but would be happy to include this in gollum `contrib`.

morus:
just to make sure I understand that correctly:
the idea is to close this PR and have a different one on gollum providing the macro there in a new folder contrib/macros. Users that want to make use of the macro would have to provide a config.rb with a line `require "contrib/macros/structured_global_toc"` (not sure how the path would look exactly, but I can figure that out).

right?

dometto:
Yes that's the current proposal!

morus:
unfortunately it turns out that this makes keeping the tests hard, as they depend on the wiki factory (could be copied, it's not a long class, but does not feel right).

Enabling the macro in that scenario isn't very nice either, you'd have to add something like
```
gollum_gem_path = Gem::Specification.find_by_name('gollum').gem_dir
require gollum_gem_path + '/contrib/macros/structured_global_toc'
```
to the config.rb as a file in gems contrib directory (outside lib) cannot be directly required.

Not sure if I want to go that way.

dometto:
On the other hand the tests in this PR do help a lot with maintainability (and not sure we can run tests on macros `contrib` at the moment, though we could also add that).

I won't have time to come back to this for a bit. Any decision about where to include this (once the test failures are resolved) is fine with me @bartkamphorst and @benjaminwil.

@morus: thanks a lot for the contributions!

morus:
Sorry, I broke the test by a tiny last minute fix. shouldn't have pushed without checking. Fixed now.

I guess tests in contrib are not a problem, but dependencies (as the wiki factory) are.

I need a brake from the topic as well.

